PR #1, opened 2019-02-09
Auth and crud module (with mongo support).

YourTechBud:
I've setup the project structure. This project will house registrator as well.

For now I've added these modules
- Auth (Supported rules: allow, deny, authenticated, match, and, or and query)
- Crud (With mongo create, read, update, support only.)


crud/mgo/read.go
```diff
@@ -0,0 +1,109 @@
...
+
+		result, err = collection.Distinct(ctx, *distinct, req.Find)
+		if err != nil {
+			return nil, err
+		}
+
+	case utils.All:
+		findOptions := options.Find()
+
+		if req.Options.Select == nil {
```
Jayesh333:
```suggestion
		if req.Options.Select != nil {
```

crud/mgo/read.go
```diff
@@ -0,0 +1,109 @@
...
+		if err := cur.Err(); err != nil {
+			return nil, err
+		}
+
+		result = results
+
+	case utils.One:
+		findOneOptions := options.FindOne()
+
+		if req.Options.Select == nil {
```
Jayesh333:
```suggestion
		if req.Options.Select != nil {
```

crud/crud.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"context"
+	"errors"
+	"sync"
+
+	"github.com/spaceuptech/space-cloud/model"
+	"github.com/spaceuptech/space-cloud/utils"
+)
+
+// Module is the root block providing convinient wrappers
```
Jayesh333:
```suggestion
// Module is the root block providing convenient wrappers
```

auth/errors.go
```diff
@@ -0,0 +1,12 @@
+package auth
+
+import "errors"
+
+// ErrRuleNotFound is thrown when an error is not present in the Auth object
+var ErrRuleNotFound = errors.New("Auth: The no rule has been provided")
```
Jayesh333:
```suggestion
var ErrRuleNotFound = errors.New("Auth: No rule has been provided")
```

PR #2, opened 2019-02-10
Sql support to crud

YourTechBud:
I'm a bit skeptical about the sql update. Have a look at that.


crud/sql/update.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/spaceuptech/space-cloud/utils"
+)
+
+// Update updates the document(s) which match the condition provided.
+func (s *SQL) Update(ctx context.Context, project, col string, req *model.UpdateRequest) error {
+	// Generate a prepared query builder
+	query := goqu.From(col).Prepared(true)
+	query = query.SetAdapter(goqu.NewAdapter(s.dbType, query))
+
+	if req.Find == nil {
```
Jayesh333:
```suggestion
	if req.Find != nil {
```

crud/sql/sql.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	s.client = sql
+	return s, nil
+}
+
+// GetDBType returns the dbType of the crud block
+func (s *SQL) GetDBType() utils.DBType {
+	switch s.dbType {
+	case "postgres":
```
Jayesh333:
```suggestion
	case "sql-postgres":
```

crud/sql/sql.go
```diff
@@ -0,0 +1,68 @@
...
+		return utils.Postgres
+	case "mysql":
```
Jayesh333:
```suggestion
	case "sql-mysql":
```

crud/sql/sql.go
```diff
@@ -0,0 +1,68 @@
...
+		return utils.MySQL
+	}
+
+	return utils.MySQL
```
Jayesh333: Shouldn't this return an error if no matching db found

crud/sql/sql.go
```diff
@@ -0,0 +1,68 @@
...
+	s := &SQL{}
+
+	switch dbType {
+	case utils.Postgres:
+		sql, err = sqlx.Open("postgres", connection)
+		s.dbType = "postgres"
+
+	case utils.MySQL:
+		sql, err = sqlx.Open("mysql", connection)
+		s.dbType = "mysql"
```
Jayesh333:
```suggestion
		s.dbType = "sql-mysql"
```

crud/sql/sql.go
```diff
@@ -0,0 +1,68 @@
...
+func Init(dbType utils.DBType, connection string) (*SQL, error) {
+	var sql *sqlx.DB
+	var err error
+
+	s := &SQL{}
+
+	switch dbType {
+	case utils.Postgres:
+		sql, err = sqlx.Open("postgres", connection)
+		s.dbType = "postgres"
```
Jayesh333:
```suggestion
		s.dbType = "sql-postgres"
```

crud/sql/sql.go
YourTechBud: its supposed to be without the "sql-" prefix. SQL.dbType is used directly in goqu

crud/sql/sql.go
YourTechBud: its supposed to be without the "sql-" prefix. SQL.dbType is used directly in goqu

crud/sql/sql.go
YourTechBud: It isn't possible that SQL.dbType would have any value apart from these two.

PR #41, opened 2019-02-28
Added grpc and some minor changes in Http and crud(model)

KhushalPShah:


grpc_server.go
```diff
@@ -0,0 +1,316 @@
...
+
+func (s *server) Create(ctx context.Context, in *pb.CreateRequest) (*pb.Response, error) {
+
+	out := pb.Response{}
+
+	authObj, err := s.auth.IsAuthenticated(in.Meta.Token, in.Meta.DBType, in.Meta.Col, utils.Create)
+	if err != nil {
+		out.Status = 401
+		out.Error = err.Error()
+		return &out, nil
```
YourTechBud: why dont you make the pb.Response object right here?

YourTechBud: You can de the same everywhere you are supposed to return a reponse object

grpc_server.go
```diff
@@ -0,0 +1,316 @@
...
+
+	// Give positive acknowledgement
+	out.Status = 200
+	out.Result = temp1
+	return &out, nil
+
+}
+
+//Rename the Function name.
+func (s *server) initGRPCServer(port string) {
```
YourTechBud: Make this function in server.go

PR #64, opened 2019-03-19
Update proto/server.proto for JAVA

AliabbasMerchant:
Do we need to put these changes in this repo also, or should we keep the changes in the JAVA repo only?
@YourTechBud @KhushalPShah


YourTechBud:
We should keep the final proto file in this repo itself.

PR #69, opened 2019-03-21
Implemented Transactions on Crud

atharva29:

### DIFF
--- a/http.go
+++ b/http.go
@@ -370,3 +370,186 @@ func getRequestMetaData(r *http.Request) *requestMetaData {
 
 	return &requestMetaData{project: project, dbType: dbType, col: col, token: tokens[0]}
 }
+
+func (s *server) handleTransaction() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// Create a context of execution
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		// Get the path parameters
+		meta := getRequestMetaData(r)
+
+		// Load the request from the body
+		var txRequest model.TransactionRequest
+		json.NewDecoder(r.Body).Decode(&txRequest)
+		defer r.Body.Close()
+
+		args := map[string]interface{}{}
+		for _, req := range txRequest.Requests {
+
+			switch req.Type {
+			case string(utils.Update):
+				authObj, err := s.auth.IsAuthenticated(meta.token, meta.dbType, req.Col, utils.Update)
+				if err != nil {
+					w.WriteHeader(http.StatusUnauthorized)
+					json.NewEncoder(w).Encode(map[string]string{"error": "You are not authenticated"})
+					return
+				}
+				args = map[string]interface{}{
+					"args":    map[string]interface{}{"find": req.Find, "update": req.Update, "op": req.Operation, "auth": authObj},
+					"project": meta.project, // Don't forget to do this for every request
+				}
+
+				// Check if user is authorized to make this request
+				err = s.auth.IsAuthorized(meta.dbType, req.Col, utils.Update, args)
+				if err != nil {
+					w.WriteHeader(http.StatusForbidden)
+					json.NewEncoder(w).Encode(map[string]string{"error": "You are not authorized to make this request"})
+					return
+				}
+
+			case string(utils.Create):
+				authObj, err := s.auth.IsAuthenticated(meta.token, meta.dbType, req.Col, utils.Create)
+				if err != nil {
+					w.WriteHeader(http.StatusUnauthorized)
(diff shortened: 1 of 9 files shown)


http.go
```diff
@@ -370,3 +370,190 @@ func getRequestMetaData(r *http.Request) *requestMetaData {
...
+					if err != nil {
+						w.WriteHeader(http.StatusUnauthorized)
+						json.NewEncoder(w).Encode(map[string]string{"error": "You are not authenticated"})
+						return
+					}
+					args = map[string]interface{}{
+						"args":    map[string]interface{}{"find": req.Find, "update": req.Update, "op": req.Operation, "auth": authObj},
+						"project": meta.project, // Don't forget to do this for every request
+					}
+					break
```
YourTechBud: In Golang you don't really need to use break in switch case. Lets remove that.

http.go
```diff
@@ -370,3 +370,190 @@ func getRequestMetaData(r *http.Request) *requestMetaData {
...
+				}
+			case "create":
```
YourTechBud: Can we use the constants in utils instead of a string literal?

modules/crud/mgo/transaction.go
```diff
@@ -8,38 +8,38 @@ import (
...
 			return err
 		}
-		for _, req := range reqs {
-			col := req["col"].(string)
+		for _, req := range reqs.Requests {
+			col := req.Col
 
-			switch req["type"].(string) {
+			switch req.Type {
 			case "create":
```
YourTechBud: Constants instead of string literals again!

modules/crud/sql/transaction.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/spaceuptech/space-cloud/model"
+)
+
+func (s *SQL) Transaction(ctx context.Context, project string, reqs *model.TransactionRequest) error {
+
+	var opts *sql.TxOptions
```
YourTechBud: Is opts supposed to be empty?

modules/crud/sql/transaction.go
```diff
@@ -0,0 +1,65 @@
...
+	tx, err := s.client.BeginTxx(ctx, opts)
+	if err != nil {
+		fmt.Println("Error in initiating transactions")
+		return err
+	}
+	for _, req := range reqs.Requests {
+		switch req.Type {
+		case "create":
```
YourTechBud: Use constants from utils instead of string literal

modules/crud/sql/transaction.go
```diff
@@ -0,0 +1,65 @@
...
+	"database/sql"
+	"fmt"
+
+	"github.com/spaceuptech/space-cloud/model"
+)
+
+func (s *SQL) Transaction(ctx context.Context, project string, reqs *model.TransactionRequest) error {
+
+	var opts *sql.TxOptions
+	tx, err := s.client.BeginTxx(ctx, opts)
```
YourTechBud: There are no tx rollbacks happening in case of errors. Lets do that.

routes.go
```diff
@@ -19,6 +19,9 @@ func (s *server) routes() {
 	crudRouter.HandleFunc("/update", s.handleUpdate())
 	crudRouter.HandleFunc("/delete", s.handleDelete())
 	crudRouter.HandleFunc("/aggr", s.handleAggregate())
+	crudRouter.HandleFunc("/transaction", s.handleTransaction())
```
YourTechBud: Why are we registering the same handler on two separate routes?

YourTechBud:
Curious to know if this is tested?

modules/crud/sql/transaction.go
atharva29: Opts has two fields IsolationLevel and one boolean ReadOnly 
What value should i give ?

modules/crud/sql/transaction.go
atharva29: Sqlx automatically handels rollback function 
 " If the context is canceled, the sql package will roll back the transaction automatically " --> GODOC

modules/crud/sql/transaction.go
YourTechBud: Turns out we can pass `nil` instead of options to use the defaults. Can we explicitly put nil and remove that opts variable. Probably add a todo comment saying opts need to be created?

modules/crud/sql/transaction.go
YourTechBud: You are right!

modules/crud/sql/transaction.go
atharva29: Yes nil can be passed . Added todo

Issue #73, opened 2019-03-22
Use context in SQL 


Jayesh333:
Use the context object in every SQL db operation similar to mongo

PR #75, opened 2019-03-22
Use context in SQL #73

u5surf:
resolve: #73 
adding context in SQL


YourTechBud:
Thanks for the pull request. Really appreciate it mahn. Can you do the same for SQL transactions as well? I think we merged it recently.

Jayesh333:
You  missed line 90 in read.go

PR #77, opened 2019-03-23
added test files for sql

sharadregoti:
Code coverage : 34.3%
some bug in the program, not able to find out.
test passes most of the time then test fails relacing string1 with string2 for update and delete test files

### DIFF
--- a/modules/crud/sql/create.go
+++ b/modules/crud/sql/create.go
@@ -16,7 +16,7 @@ import (
 
 // Create inserts a document (or multiple when op is "all") into the database
 func (s *SQL) Create(ctx context.Context, project, col string, req *model.CreateRequest) error {
-	sqlQuery, args, err := s.generateCreateQuery(ctx, project, col, req)
+	sqlQuery, args, err := s.GenerateCreateQuery(ctx, project, col, req)
 	if err != nil {
 		return err
 	}
@@ -24,7 +24,7 @@ func (s *SQL) Create(ctx context.Context, project, col string, req *model.Create
 }
 
 //generateCreateQuery makes query for create operation
-func (s *SQL) generateCreateQuery(ctx context.Context, project, col string, req *model.CreateRequest) (string, []interface{}, error) {
+func (s *SQL) GenerateCreateQuery(ctx context.Context, project, col string, req *model.CreateRequest) (string, []interface{}, error) {
 	// Generate a prepared query builder
 	query := goqu.From(col).Prepared(true)
 	query = query.SetAdapter(goqu.NewAdapter(s.dbType, query))
--- a/modules/crud/sql/delete.go
+++ b/modules/crud/sql/delete.go
@@ -14,7 +14,7 @@ import (
 
 // Delete removes the document(s) from the database which match the condition
 func (s *SQL) Delete(ctx context.Context, project, col string, req *model.DeleteRequest) error {
-	sqlString, args, err := s.generateDeleteQuery(ctx, project, col, req)
+	sqlString, args, err := s.GenerateDeleteQuery(ctx, project, col, req)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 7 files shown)


modules/crud/sql/delete_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/spaceuptech/space-cloud/model"
+)
+
+func TestGenerateDeleteQuery(t *testing.T) {
+	// truecases := 11
+	tests := []struct {
+		name, tableName, want string
+		req                   model.DeleteRequest
+	}{
+		{name: "Successfull Test", tableName: "fooTable", want: "DELETE FROM fooTable WHERE ((String1 = ?) AND (String2 = ?))", req: model.DeleteRequest{Find: map[string]interface{}{"String1": "1", "String2": "2"}}},
```
sharadregoti: output should be this : "DELETE FROM fooTable WHERE ((String1 = ?) AND (String2 = ?))"
but sometimes string1 & string2 interchange
"DELETE FROM fooTable WHERE ((String2 = ?) AND (String1 = ?))"

modules/crud/sql/update_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func TestGenerateUpdateQuery(t *testing.T) {
+	truecases := 10
+	var ctx context.Context
+	project := "projectName"
+	tests := []struct {
+		name, tableName, want string
+		req                   model.UpdateRequest
+	}{
+		{name: "Successfull Test", tableName: "fooTable", want: "UPDATE fooTable SET String1=? WHERE ((FindString1 = ?) AND (FindString2 = ?))", req: model.UpdateRequest{Update: map[string]interface{}{"$set": map[string]interface{}{"String1": "1"}}, Find: map[string]interface{}{"FindString1": "1", "FindString2": "2"}}},
```
sharadregoti: same here

modules/crud/sql/delete_test.go
Jayesh333: The where clause parameters may get interchanged. Write your test cases to accommodate both the results using the or

modules/crud/sql/update_test.go
Jayesh333: Same

modules/crud/sql/create.go
```diff
@@ -16,15 +16,15 @@ import (
 
 // Create inserts a document (or multiple when op is "all") into the database
 func (s *SQL) Create(ctx context.Context, project, col string, req *model.CreateRequest) error {
-	sqlQuery, args, err := s.generateCreateQuery(ctx, project, col, req)
+	sqlQuery, args, err := s.GenerateCreateQuery(ctx, project, col, req)
```
Jayesh333: Don't export this function

modules/crud/sql/create.go
```diff
@@ -16,15 +16,15 @@ import (
...
 	if err != nil {
 		return err
 	}
 	return s.doExec(sqlQuery, args)
 }
 
 //generateCreateQuery makes query for create operation
-func (s *SQL) generateCreateQuery(ctx context.Context, project, col string, req *model.CreateRequest) (string, []interface{}, error) {
+func (s *SQL) GenerateCreateQuery(ctx context.Context, project, col string, req *model.CreateRequest) (string, []interface{}, error) {
```
Jayesh333: don't export this function

modules/crud/sql/delete.go
```diff
@@ -14,15 +14,15 @@ import (
...
+	sqlString, args, err := s.GenerateDeleteQuery(ctx, project, col, req)
 	if err != nil {
 		return err
 	}
 	return s.doExec(sqlString, args)
 }
 
 //genrateDeleteQuery makes query for delete operation
-func (s *SQL) generateDeleteQuery(ctx context.Context, project, col string, req *model.DeleteRequest) (string, []interface{}, error) {
+func (s *SQL) GenerateDeleteQuery(ctx context.Context, project, col string, req *model.DeleteRequest) (string, []interface{}, error) {
```
Jayesh333: don't export this function

modules/crud/sql/delete_test.go
Jayesh333: put an or clause. both the results are right

modules/crud/sql/update_test.go
Jayesh333: both results are right

modules/crud/sql/create.go
sharadregoti: I tried to not export these func
But while running go test command
It says not test files found or undefined function name

YourTechBud: so the name of the testing function in *_test.go will always start with "Test". So the test functions will always be exported but the function to be tested can be un-exported.

PR #94, opened 2019-03-30
Refactor security docs

ayan4m1:
Two changes here:

* Make the console message printed on doc server startup a valid URL (e.g. `Documentation served from http://192.168.0.1:3000/docs`)
* Reword and expand some parts of the security overview doc

Happy to accept change requests. Thanks for your time.

### DIFF
--- a/docs/index.js
+++ b/docs/index.js
@@ -86,4 +86,4 @@ app.get('/docs/:dir/:file', handlePage)
 
 
 const port = process.env.POST | 3000
-app.listen(port, () => console.log(`Example app listening on port ${port}!`))
+app.listen(port, () => console.log(`Documentation served from http://localhost:${port}/docs`))
--- a/docs/manual/security/index.md
+++ b/docs/manual/security/index.md
@@ -4,7 +4,7 @@ Space Cloud leverages multiple security systems with the philosophy that applica
 
 Every module in Space Cloud supports JWT-based authentication. Access to any database not specified in the configuration file is denied by default. Whenever possible, use JWT-based authentication in production. You can check out the [official website](https://jwt.io) of the JWT project to learn more about it.
 
-## How JWT are tokens used
+## How JWT is used
 
 `space-cloud` looks for the JWT token in the `Authorization` header in each HTTP request made to it.
 
--- a/docs/manual/security/index.md
+++ b/docs/manual/security/index.md
@@ -2,7 +2,7 @@
 
 Space Cloud leverages multiple security systems with the philosophy that applications and platforms should be secure by default. Treating security as an afterthought is a recipe for disaster.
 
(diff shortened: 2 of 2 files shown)


Jayesh333:
Thanks buddy @ayan4m1 for the pull request. Really appreciate it. The localhost in the console message printed on docs server can be replaced by more technically correct address - 0.0.0.0 though. Can you update the pull request and squash the commits into one? @ayan4m1 Rest everything looks good.

YourTechBud:
> It is disabled by default. This is because access to any database not specified in the configuration file is denied by default. Whenever possible, enable JWT-based authentication in production.

Thinking about it, this too is incorrect. Jwt can never be disabled. You can switch the rule to `allow` to bypass the authentication and authorization mechanism for that particular operation on that particular table.

We should probably remove the line that it's disabled by default. Sorry for the confusion!

It would be great if you could squash you commits into one!

ayan4m1:
Sure, I'll make the updates - you should be able to choose "Squash and merge" from within GitHub but I can make a single commit if you need it.

ayan4m1:
@Jayesh333 I added the [ip]() package in order to display the first available IP address of the system instead of `localhost` - I wanted the link in the console to be clickable and the browser wouldn't know what to do with `http://0.0.0.0`

YourTechBud:
@Jayesh333 are you okay with introducing a new library for detecting the ip address or should we stick to `localhost` ?

Jayesh333:
For those who are running docs locally, `localhost` would work perfectly. And for those watching docs online, it isn't a problem at all. So I think `localhost` will be fine. @YourTechBud . Let's stick to `localhost`

docs/index.js
```diff
@@ -84,6 +84,6 @@ app.get('/docs/sitemap.xml', handleSitemap)
 app.get('/docs/:dir', (req, res) => res.redirect(`/docs/${req.params.dir}/overview`))
 app.get('/docs/:dir/:file', handlePage)
 
-
 const port = process.env.POST | 3000
-app.listen(port, () => console.log(`Example app listening on port ${port}!`))
\ No newline at end of file
+const proto = process.env.SECURE ? 'https' : 'http'
+app.listen(port, () => console.log(`Documentation served from ${proto}://${ip.address()}:${port}/docs`))
```
YourTechBud: Stick to localhost. Let's get rid of the ip library.

docs/package.json
```diff
@@ -13,6 +13,7 @@
   "dependencies": {
     "express": "^4.16.4",
     "handlebars": "^4.1.0",
+    "ip": "^1.1.5",
```
YourTechBud: And let's make sure we remove it from here as well

ayan4m1:
I just removed the single commit that contained all of the `ip` related changes. It now shows `localhost`.

PR #99, opened 2019-04-01
added test files for auth module

sharadregoti:
added test file for auth module
code coverage 74%
matchQuery func is still not tested


YourTechBud:
Bump the minor version up by 1 since you have fixed a bug as well. Make it 6.0.2!

PR #106, opened 2019-04-03
Refactored from Transaction to Batch

KhushalPShah:


modules/crud/sql/sql.go
```diff
@@ -87,7 +87,7 @@ func (s *SQL) doExecContext(ctx context.Context, query string, args []interface{
 	return err
 }
 
-func doTransactionExecContext(ctx context.Context, query string, args []interface{}, tx *sqlx.Tx) error {
+func doBatchExecContext(ctx context.Context, query string, args []interface{}, tx *sqlx.Tx) error {
```
YourTechBud: Let this function be `doTransactionExecContext`

PR #113, opened 2019-04-05
fixed bug in file auth

atharva29:


Jayesh333:
You have the config file of example app. Also the code is not indented properly

Issue #135, opened 2019-04-14
No scrollbar visible in the docs pages


AliabbasMerchant:
Problem exists in Mozilla Firefox, Google Chrome  
OS: Windows 10

YourTechBud:
@akshayjj1997 can you have a look at it?

YourTechBud:
Fixed in #138. Thanks @nicknonis for the super quick fix!

PR #136, opened 2019-04-14
No scrollbar visible in the docs pages [Patch]

nicknonis:
created initial request here: https://github.com/YourTechBud/space-cloud/pull/1
 
[docs/public/docs/sass/index/_content.scss]
changes made to .main-wrapper {.content-wrapper{ 
-left property change to 0
-padding property changed to 40px 0px 0px 360px;

[docs/public/docs/sass/index/_sidebar.scss]
added a z-index of 1 to .main-wrapper >  .desktop-docs-sidebar

### DIFF
--- a/docs/public/docs/sass/index/_content.scss
+++ b/docs/public/docs/sass/index/_content.scss
@@ -6,11 +6,11 @@
     @media #{$large-and-up} {
       position: absolute;
       top: 64px;
-      left: 300px;
+      left: 0;
       right: 0;
       bottom: 0;
       overflow: auto;
-      padding: 40px 0px 0px 60px;
+      padding: 40px 0px 0px 360px;
     }
 
     .docs-content {
--- a/docs/public/docs/sass/index/_sidebar.scss
+++ b/docs/public/docs/sass/index/_sidebar.scss
@@ -8,6 +8,7 @@
     background: $default-color;
     border-right: 1px solid $border-color;
     overflow: auto;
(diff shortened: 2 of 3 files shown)


YourTechBud:
This fixes #135. Can you squash all the commits into one? @akshayjj1997 quickly have a look at this and approve if everything looks good!

YourTechBud:
> Problem exists in Mozilla Firefox, Google Chrome
OS: Windows 10

See if this is fixed in these environments as well!

akshayjj1997:
The content feels like stacked to the left. Needs to change the class="col s12 l6" of "content-wrapper" inside index.handlebars to s12 l7 or s12 l8

YourTechBud:
Yeah the width of the content seems to be a bit too packed now. @akshayjj1997 do we even need a grid there? Why not have a grid inside the div to which the padding is applied?

nicknonis:
I think I've squashed the commits, sorry first time pull request :) 
Adding a grid to the div sounds like an improvement.

akshayjj1997:
@YourTechBud Yes we need a grid there, otherwise the entire content will take the full width. And the gird is placed inside the div to which the padding is applied

YourTechBud:
No problem @nicknonis! Glad to be your first contribution! Can you make the necessary changes suggested by @akshayjj1997?

nicknonis:
okay done! is it any better now?

PR #142, opened 2019-04-14
Fix typo

shyim:


YourTechBud:
I really appreciate the pull request.

What want to show that we support Postgres as well as the databases compatible with Postgres.

I have a feeling we are changing the underlying meaning here.

PR #151, opened 2019-04-16
Real time support for GRPC

KhushalPShah:


utils/client.go
```diff
@@ -51,16 +93,38 @@ func (c *Client) Close() {
...
+				return
+			}
+			var data map[string]interface{}
+			data["Token"] = in.Token
+			data["DBType"] = in.DbType
+			data["Project"] = in.Project
+			data["Group"] = in.Group
+			data["Type"] = in.Type
+			data["ID"] = in.Id
+			data["Where"] = in.Where
```
YourTechBud: i guess `in.Where` is a []byte. The `data["where"]` must be a map[string]interface{} i guess. You should Unmarshal it first.

utils/client.go
```diff
@@ -70,8 +134,16 @@ func (c *Client) ClientID() string {
 
 // CreateWebsocketClient makes a client object to manage the socket
 func CreateWebsocketClient(socket *websocket.Conn) *Client {
-	channel := make(chan interface{}, 5)
+	channel := make(chan *model.Message, 5)
+	ctx, cancel := context.WithCancel(context.Background())
+	id := uuid.NewV1().String()
+	return &Client{id, channel, ctx, cancel, socket, nil, "WEBSOCKET"}
```
YourTechBud: can we make the `WEBSOCKET` a constant in the utils package?

utils/client.go
```diff
@@ -70,8 +134,16 @@ func (c *Client) ClientID() string {
...
+}
+
+// CreateGRPCClient makes a client object to manage the grpc
+func CreateGRPCClient(stream pb.SpaceCloud_RealTimeServer) *Client {
+	channel := make(chan *model.Message, 5)
 	ctx, cancel := context.WithCancel(context.Background())
 	id := uuid.NewV1().String()
-	return &Client{id, channel, ctx, cancel, socket}
+	return &Client{id, channel, ctx, cancel, nil, stream, "GRPC"}
```
YourTechBud: can we make the `GRPC` a constant in the utils package.

utils/client.go
```diff
@@ -51,16 +93,38 @@ func (c *Client) Close() {
...
 // Read startes a blocking reader routine
 func (c *Client) Read(cb DataCallback) {
-	defer c.Close()
-	for {
-		data := &model.Message{}
-		err := c.socket.ReadJSON(data)
-		if err != nil {
-			return
+	if c.protocol == "WEBSOCKET" {
+		defer c.Close()
```
YourTechBud: can we put differ on the top just once?

utils/client.go
```diff
@@ -2,40 +2,82 @@ package utils
...
+					grpcResponse := pb.RealTimeResponse{Id: res.ID, Group: responseMsg.Group, Ack: responseMsg.Ack, Error: responseMsg.Error, FeedData: feedData}
+					c.stream.Send(&grpcResponse)
+
+				case TypeRealtimeFeed:
+					feed := res.Data.(model.FeedData)
+					var feedData []*pb.FeedData
+					payload, err := json.Marshal(feed.Payload)
+					if err != nil {
+						grpcResponse := pb.RealTimeResponse{Error: err.Error()}
+						c.stream.Send(&grpcResponse)
```
YourTechBud: Just log the error. Dont stream errors to the client. Also you need to return straight after logging the error.

utils/client.go
```diff
@@ -2,40 +2,82 @@ package utils
...
+							c.stream.Send(&grpcResponse)
+						}
+						feedData[i] = &pb.FeedData{QueryId: feed.QueryID, DocId: feed.DocID, Type: feed.Type, Group: feed.Group, DbType: feed.DBType, Payload: payload, TimeStamp: feed.TimeStamp}
+					}
+					grpcResponse := pb.RealTimeResponse{Id: res.ID, Group: responseMsg.Group, Ack: responseMsg.Ack, Error: responseMsg.Error, FeedData: feedData}
+					c.stream.Send(&grpcResponse)
+
+				case TypeRealtimeFeed:
+					feed := res.Data.(model.FeedData)
+					var feedData []*pb.FeedData
```
YourTechBud: feedData should be defined and declared using something like
```go
feedData := make([]*pb.FeedData, 1)
```

YourTechBud:
```suggestion
					feedData := make([]*pb.FeedData, 1)
```

utils/client.go
```diff
@@ -52,15 +94,41 @@ func (c *Client) Close() {
 // Read startes a blocking reader routine
 func (c *Client) Read(cb DataCallback) {
 	defer c.Close()
-	for {
-		data := &model.Message{}
-		err := c.socket.ReadJSON(data)
-		if err != nil {
-			return
+	if c.protocol == Websocket {
```
YourTechBud: Can we use a switch case here?

utils/client.go
```diff
@@ -2,40 +2,82 @@ package utils
...
 type DataCallback func(data *model.Message)
 
 // RoutineWrite starts a json writer routine
 func (c *Client) RoutineWrite() {
-	go func() {
-		for res := range c.channel {
-			err := c.socket.WriteJSON(res)
-			if err != nil {
-				log.Println(err)
+	if c.protocol == Websocket {
```
YourTechBud: Can we use a switch case here

utils/client.go
```diff
@@ -2,40 +2,83 @@ package utils
...
-	}()
+		}()
+	case GRPC:
+		go func() {
+			for res := range c.channel {
+				//Convert the Message into RealTime response.
+				switch res.Type {
+				case TypeRealtimeSubscribe, TypeRealtimeUnsubscribe:
+					//Decode the Message
+					feedData := make([]*pb.FeedData, 1)
```
YourTechBud:
```suggestion
					feedData := make([]*pb.FeedData, len(responseMsg.Docs))
```

utils/client.go
```diff
@@ -52,14 +95,42 @@ func (c *Client) Close() {
...
+			data["Token"] = in.Token
+			data["DBType"] = in.DbType
+			data["Project"] = in.Project
+			data["Group"] = in.Group
+			data["Type"] = in.Type
+			data["ID"] = in.Id
+			var temp interface{}
+			err = json.Unmarshal(in.Where, &temp)
+			if err != nil {
+				return
```
YourTechBud: Let's log the error here as well!

utils/client.go
YourTechBud: you'll have to move line number 49 above 48

YourTechBud:
Also, you've got to resolve the merge conflict as well!

utils/client.go
KhushalPShah: How can I get the Length of the Docs?
Cos that will correspondingly be the Length of my array of pb.FeedData.

KhushalPShah: Got it..Thanks!

PR #161, opened 2019-04-17
Rewrite if-else to switch statements

sosiska:
This commit rewrites some if-else-if-else chains as a switch.
Based on Go style guide:
https://golang.org/doc/effective_go.html#switch

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 	<strong>
 		<a href="https://spaceuptech.com/">Website</a>
 		•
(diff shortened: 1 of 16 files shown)


YourTechBud:
Can you change the base to `0.7.0` branch instead?

sosiska:
> Can you change the base to `0.7.0` branch instead?
If I understood correctly, here you are:
https://github.com/spaceuptech/space-cloud/pull/162

YourTechBud:
Done in #162

PR #162, opened 2019-04-17
Rewrite if-else to switch statements

sosiska:
This commit rewrites some if-else-if-else chains as a switch.
Based on Go style guide:
https://golang.org/doc/effective_go.html#switch


YourTechBud:
Perfect. Merging this now. Really appreciate the contribution!

PR #164, opened 2019-04-18
Change connection-string for SQL-databases

### LINKED ISSUE #158: Replace "root" in default connection string for SQL databases
Currently the docs and space-cloud init default to using the root user in the MySQL connection string. It would be better to use a different name like "user" to guide users away from actually connecting with the root user. Something like this:
```yaml
conn: "user:my-secret-pwd@/project"
```

ledex:
Change connection-string for SQL-databases from "root:my-secret-pw@/test" to "user:my-secret-pwd@/project", so that login with the root account by default is prevented. This fixes #158.

### DIFF
--- a/config/generate.go
+++ b/config/generate.go
@@ -88,7 +88,7 @@ func getConnectionString(db string) string {
 	case string(utils.Mongo):
 		return "mongodb://localhost:27017"
 	case string(utils.MySQL):
-		return "root:my-secret-pw@/test"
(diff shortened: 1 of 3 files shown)


YourTechBud:
Since this modifies the source code can you bump the version (patch) number up by one. Also it would be crazy if you could also fix this in the docs here: https://spaceuptech.com/docs/database/config

ledex:
Should be good now.

PR #182, opened 2019-04-22
Fix/84 unsave password handling

### LINKED ISSUE #84: unsafe password handling
actual plaintext passwords being sent in user management, really unsafe thing to do. Standard way is to salt the passwords and the compare the hashes, not plain text passwords

ledex:
Add hashing and salting of passwords in the database with [bcrypt](https://godoc.org/golang.org/x/crypto/bcrypt). This fixes #84. 

Added that to the next minor version, because changes to the database have to be made and old users have to be transferred.

### DIFF
--- a/modules/userman/email.go
+++ b/modules/userman/email.go
@@ -7,6 +7,8 @@ import (
 	"net/http"
 	"time"
 
+	"golang.org/x/crypto/bcrypt"
+
 	"github.com/gorilla/mux"
 	uuid "github.com/satori/go.uuid"
 
@@ -40,7 +42,7 @@ func (m *Module) HandleEmailSignIn() http.HandlerFunc {
 		defer r.Body.Close()
 
 		// Create read request
-		readReq := &model.ReadRequest{Find: map[string]interface{}{"email": req["email"], "pass": req["pass"]}, Operation: utils.One}
+		readReq := &model.ReadRequest{Find: map[string]interface{}{"email": req["email"]}, Operation: utils.One}
 
 		user, err := m.crud.Read(ctx, dbType, project, "users", readReq)
 		if err != nil {
@@ -49,14 +51,27 @@ func (m *Module) HandleEmailSignIn() http.HandlerFunc {
 			return
 		}
 
-		// Delete password from user
 		userObj := user.(map[string]interface{})
+
+		//Compares if the given password is correct
+		err = bcrypt.CompareHashAndPassword([]byte(userObj["pass"].(string)), []byte(req["pass"].(string)))
+		if err != nil {
+			w.WriteHeader(http.StatusUnauthorized)
+			json.NewEncoder(w).Encode(map[string]string{"error": "Given credentials are not correct"})
+			return
+		}
+
+		// Delete password from user
 		delete(userObj, "pass")
 		delete(req, "pass")
 
 		// Create a token
+		if dbType == "mongo" {
+			req["_id"] = userObj["_id"]
+		} else {
+			req["id"] = userObj["id"]
+		}
 		req["role"] = userObj["role"]
-		req["name"] = userObj["name"]
 		token, err := m.auth.CreateToken(req)
 		if err != nil {
 			w.WriteHeader(http.StatusInternalServerError)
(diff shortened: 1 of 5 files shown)


ledex:
Currently, the password is still represented as a string. This might not be wanted and instead, the password should be represented as binary.

modules/userman/email.go
```diff
@@ -49,8 +51,17 @@ func (m *Module) HandleEmailSignIn() http.HandlerFunc {
...
+		err = bcrypt.CompareHashAndPassword([]byte(userObj["pass"].(string)), []byte(req["pass"].(string)))
+		if err != nil {
+			w.WriteHeader(http.StatusUnauthorized)
+			json.NewEncoder(w).Encode(map[string]string{"error": "Given credentials are not correct"})
+			return
+		}
+
+		// Delete password from user
 		delete(userObj, "pass")
 		delete(req, "pass")
```
Jayesh333: No need to delete `pass` from request

modules/userman/email.go
```diff
@@ -132,3 +146,18 @@ func (m *Module) HandleEmailSignUp() http.HandlerFunc {
 		json.NewEncoder(w).Encode(map[string]interface{}{"user": req, "token": token})
 	}
 }
+
+func hashPassword(pwd string) string {
+	//Generates a new hash from the given password
+	hash, err := bcrypt.GenerateFromPassword([]byte(pwd), bcrypt.MinCost)
+	if err != nil {
+		panic(err.Error)
```
Jayesh333: Can we return an InternalServerError instead of panic?

modules/userman/email.go
```diff
@@ -132,3 +146,18 @@ func (m *Module) HandleEmailSignUp() http.HandlerFunc {
...
+	}
+
+	//Checks if the hash is correct for the given password
+	err = bcrypt.CompareHashAndPassword(hash, []byte(pwd))
+	if err != nil {
+		panic(err.Error)
```
Jayesh333: Return an error rather than panic

modules/userman/email.go
ledex: Sure, but I would suggest doing both so that the administrator can see in the logs that an error was raised.

ledex:
> Need to add `id` / `_id` (for SQL and MongoDb respectively) in the `req` object from the `userObj` before creating `token` in the signin as it is done for signup

Done 👍

modules/userman/email.go
```diff
@@ -49,12 +51,26 @@ func (m *Module) HandleEmailSignIn() http.HandlerFunc {
...
+			return
+		}
+
+		// Delete password from user
 		delete(userObj, "pass")
 		delete(req, "pass")
 
 		// Create a token
+		if dbType == "mongo" {
+			req["_id"] = userObj["_id"]
```
YourTechBud:
```suggestion
			req["id"] = userObj["_id"]
```

modules/userman/email.go
```diff
@@ -49,12 +51,26 @@ func (m *Module) HandleEmailSignIn() http.HandlerFunc {
...
+			json.NewEncoder(w).Encode(map[string]string{"error": "Given credentials are not correct"})
+			return
+		}
+
+		// Delete password from user
 		delete(userObj, "pass")
 		delete(req, "pass")
 
 		// Create a token
+		if dbType == "mongo" {
```
YourTechBud:
```suggestion
		if dbType == string(utils.Mongo) {
```

modules/userman/email.go
```diff
@@ -49,12 +51,26 @@ func (m *Module) HandleEmailSignIn() http.HandlerFunc {
...
+			req["_id"] = userObj["_id"]
+		} else {
+			req["id"] = userObj["id"]
+		}
 		req["role"] = userObj["role"]
 		req["name"] = userObj["name"]
```
YourTechBud: We don't need the name to be there in the token. Will increase the token size.

modules/userman/email.go
```diff
@@ -94,9 +110,19 @@ func (m *Module) HandleEmailSignUp() http.HandlerFunc {
...
+		req["pass"], err = hashPassword(req["pass"].(string))
+		if err != nil {
+			log.Println("Err: ", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			json.NewEncoder(w).Encode(map[string]string{"error": "Failed to hash password"})
+			return
+		}
+
 		// Create read request
 		readReq := &model.ReadRequest{Find: map[string]interface{}{"email": req["email"], "pass": req["pass"]}, Operation: utils.One}
```
YourTechBud:
```suggestion
		readReq := &model.ReadRequest{Find: map[string]interface{}{"email": req["email"]}, Operation: utils.One}
```

YourTechBud: Can't keep the pass here since it will be different every time.

ledex:
Okey. Should be good to go... Hopefully 😄

PR #185, opened 2019-04-23
Fix/176 server url path for proxies

### LINKED ISSUE #176: Server URL Path for Proxies
I've noticed the path for the server starts with /v1/api. Assuming there could be a v2, v3, etc it would be better to have it as /api/v1, /api/v2, etc in regards to proxies. Think HA Proxy for example. If you have the version first then every time there is a new route version you'd have to HA Proxy config or have a regex matcher. While regex would work it has its own issues.

    acl container_api           path_beg        -i /v1
    acl container_api           path_beg        -i /v2
    acl container_api           path_beg        -i /v3

vs

    acl container_api           path_beg        -i /api

With the second method, you could even send different versions to different containers.

ledex:
Put the versioning behind the /api-endpoint so that when configuring a proxy only one entpoint has to be configured for multiple versions. This fixes #176.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@
     <a href="https://travis-ci.com/spaceuptech/space-cloud"><img
 		alt="Build Status"
 		src="https://travis-ci.com/spaceuptech/space-cloud.svg?branch=master"></a>
+	<a href="https://goreportcard.com/report/github.com/spaceuptech/space-cloud">
+		<img alt="Go Report Card" src="https://goreportcard.com/badge/github.com/spaceuptech/space-cloud">
+	</a>
     <br/>
     <a href="https://opensource.org/licenses/Apache-2.0"><img
 		alt="Apache License"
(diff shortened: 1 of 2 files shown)


YourTechBud:
I appreciate the contribution. Thats a breaking change. All client libraries will have to be updated in order for this to work.

@Jayesh333 We need to update the js client for this one. What do you suggest?

ledex:
Okey. But that should only be a matter of find and replace, right?

YourTechBud:
Sure. But after thinking about it, we aren't really sure if we want to do that. Or what exactly is it that we want to do. We might have to put this on hold.

ledex:
Okay. I also was not quite sure if it is best practise to not have the "base-URL" in one central place. Also, I am not sure how you want to use versioning in the endpoint.

YourTechBud:
Honestly I have no clue myself. According to me it must start with the prefix `/v1/api/:projectId`. In order to make upgrades through breaking changes seemless, the prefix `/api/:projectId/v1` seems to be the better choice.

What do you suggest?

YourTechBud:
@ledex I feel we should let the URL prefixes just be for the time being. We need to figure whats the versioning strategy to be used for the HTTP endpoints first. Closing this one.

PR #189, opened 2019-04-25
Add check for valid email #183

### LINKED ISSUE #183: Add check for valid email
When signing up a user the validity of the email address should be checked by sending an email with a confirmation-URL. Also one should be able to set if a user with an unconfirmed email is allowed to sign in.

u5surf:
fixes #183


YourTechBud:
Appreciate the contribution, but the issue has changed now. Will have to close this request!

PR #194, opened 2019-04-25
Fix/174 need versioned downloads

### LINKED ISSUE #174: Need Versioned Downloads
Long term only having the latest download is not acceptable. Need to be able to version lock ie ```wget https://spaceuptech.com/downloads/linux/space-cloud-0.6.7.zip```

edit: Better yet yarn add space-cloud@0.6.7

ledex:
Add version to deployment so that all build versions are stored. The version is determined from the ./version.go. This fixes #174.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -5,16 +5,21 @@ GOOS=linux GOARCH=amd64 go build -ldflags '-s -w -extldflags "-static"' .
 BUILD_VERSION=$(space-cloud -v | cut -f3 -d ' ')
 
 mkdir linux && mkdir windows && mkdir darwin
-zip space-cloud_v$BUILD_VERSION.zip space-cloud
-mv ./space-cloud_v$BUILD_VERSION.zip ./linux/
+zip space-cloud.zip space-cloud
+mv ./space-cloud.zip ./linux/
+cp ./linux/space-cloud.zip ./linux/space-cloud_v$BUILD_VERSION.zip 
 rm space-cloud
+
 GOOS=darwin GOARCH=amd64 go build -ldflags '-s -w -extldflags "-static"' .
-zip space-cloud_v$BUILD_VERSION.zip space-cloud
-mv ./space-cloud_v$BUILD_VERSION.zip ./darwin/
+zip space-cloud.zip space-cloud
+mv ./space-cloud.zip ./darwin/
+cp ./darwin/space-cloud.zip ./darwin/space-cloud_v$BUILD_VERSION.zip 
 rm space-cloud
+
 GOOS=windows GOARCH=amd64 go build -ldflags '-s -w -extldflags "-static"' .
-zip space-cloud_v$BUILD_VERSION.zip space-cloud.exe
-mv ./space-cloud_v$BUILD_VERSION.zip ./windows/
+zip space-cloud.zip space-cloud.exe
+mv ./space-cloud.zip ./windows/
+cp ./windows/space-cloud.zip ./windows/space-cloud_v$BUILD_VERSION.zip 
 rm space-cloud.exe
 
 # echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,22 +1,31 @@
 #!/bin/bash
 
+#Build linux-version first and determine the build-version
+GOOS=linux GOARCH=amd64 go build -ldflags '-s -w -extldflags "-static"' .
+BUILD_VERSION=$(space-cloud -v | cut -f3 -d ' ')
+
 mkdir linux && mkdir windows && mkdir darwin
+zip space-cloud_v$BUILD_VERSION.zip space-cloud
+mv ./space-cloud_v$BUILD_VERSION.zip ./linux/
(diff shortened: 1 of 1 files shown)


ledex:
I am not sure if this is the best way to determine the version. I can be exploited pretty easily. 
Does anyone have a better idea?

YourTechBud:
We can use the `./space-cloud -v` to retrieve the version. Also, i think we should keep a plain `space-cloud.zip` file along with the versioned one. So we will be uploading 6 binaries.

ledex:
@YourTechBud okay. I have changed the way the version is determined and also added the three plain ```space-cloud.zip``` files.

YourTechBud:
@ledex seems perfect to me. Let's just squash all commits into one?

@Jayesh333 You should probably have a look at it and probably test it out. You were the one to set up travis after all.

ledex:
@YourTechBud everything is squashed

YourTechBud:
@Jayesh333 test it out once. I think it looks good.

deploy.sh
```diff
@@ -1,22 +1,31 @@
...
-GOOS=linux GOARCH=amd64 go build -ldflags '-s -w -extldflags "-static"' .
-zip space-cloud.zip space-cloud
-mv ./space-cloud.zip ./linux/
-rm space-cloud
 
 # echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
 # docker push spaceuptech/space-cloud:latest
 
+#Upload as versioned build
+curl -H "Authorization: Bearer $JWT_TOKEN" -F 'file=@./darwin/space-cloud_'$BUILD_VERSION'.zip' -F 'fileType=file' -F 'makeAll=false' -F 'path=/darwin' https://spaceuptech.com/v1/api/downloads/files
```
Jayesh333: The single quotes over here are not required and might give an error:
```
'$BUILD_VERSION'.zip
```

Jayesh333:
Also, we need to prefix the $BUILD_VERSION with `v`. So the final zip should look like `space-cloud_v$BUILD_VERSION`. 

It looks good overall. Make the requested changes and we are good to go! @ledex

deploy.sh
ledex: If the single quotes are removed like so 
```
curl -H "Authorization: Bearer $JWT_TOKEN" -F 'file=@./darwin/space-cloud_$BUILD_VERSION.zip' -F 'fileType=file' -F 'makeAll=false' -F 'path=/darwin' https://spaceuptech.com/v1/api/downloads/files

```
the variable is not processed and the file `space-cloud_$BUILD_VERSION.zip` is being looked for instead.

deploy.sh
```diff
@@ -1,22 +1,31 @@
 #!/bin/bash
 
+#Build linux-version first and determine the build-version
+GOOS=linux GOARCH=amd64 go build -ldflags '-s -w -extldflags "-static"' .
+BUILD_VERSION=$(space-cloud -v | cut -f3 -d ' ')
+
+zip space-cloud_$BUILD_VERSION.zip space-cloud
+mv ./space-cloud_$BUILD_VERSION.zip ./linux/
```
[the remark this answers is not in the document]
ledex: Put the folder-creation in front now.

deploy.sh
ledex: Okay. changed it to be in double-quotes completely.

ledex:
Also squashed my changes ;) @YourTechBud

deploy.sh
```diff
@@ -1,22 +1,31 @@
...
 # echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
 # docker push spaceuptech/space-cloud:latest
 
+#Upload as versioned build
+curl -H "Authorization: Bearer $JWT_TOKEN" -F "file=@./darwin/space-cloud_v$BUILD_VERSION.zip" -F 'fileType=file' -F 'makeAll=false' -F 'path=/darwin' https://spaceuptech.com/v1/api/downloads/files
+curl -H "Authorization: Bearer $JWT_TOKEN" -F "file=@./windows/space-cloud_v$BUILD_VERSION.zip" -F 'fileType=file' -F 'makeAll=false' -F 'path=/windows' https://spaceuptech.com/v1/api/downloads/files
+curl -H "Authorization: Bearer $JWT_TOKEN" -F "file=@./linux/space-cloud_v$BUILD_VERSION.zip" -F 'fileType=file' -F 'makeAll=false' -F 'path=/linux' https://spaceuptech.com/v1/api/downloads/files
+
+#Upload as latest build
 curl -H "Authorization: Bearer $JWT_TOKEN" -F 'file=@./darwin/space-cloud.zip' -F 'fileType=file' -F 'makeAll=false' -F 'path=/darwin' https://spaceuptech.com/v1/api/downloads/files
```
Jayesh333: `space-cloud.zip` does not exists. We should copy the `space-cloud_v$BUILD_VERSION.zip` as `space-cloud.zip` before this step.

Jayesh333:
This is the last change @ledex

deploy.sh
ledex: Okey should finally be good to go :)

Issue #150, opened 2019-04-16
Support reading secrets from secure files


ollykel:
Currently the config file format requires you to provide login credentials for SQL database connections like this:
```yaml
conn: "root:my-secret-pwd@/project"
```
This exposes the user's login credentials to anyone who gets access to the source code or the image.

Services like [Docker Secrets](https://docs.docker.com/engine/swarm/secrets/) avoid this by storing secrets separately from the image and providing them to the container at startup. To support this, space-cloud should provide the option to read secrets from a file. For example, instead of the conn option provided above, users should have the option to provide a "conn_file":
```yaml
conn_file: /run/secrets/mysqlconn.txt
```
Space-cloud should check that at least one option is present in the config file at startup. If both options are present it should log a fatal error.

YourTechBud:
Thats sounds like an interesting idea! Another way we could do this is provide the secrets after space cloud has started via a configurator service. That would also help us sync the config of the various space clouds (in a Distributed environment).

But lets start by reading the secrets from a separate file. We will require a standardized format for that though.

Jayesh333:
Would reading these secrets through environment variable be a good idea?

YourTechBud:
Oh we can have an environment variable mentioned in the config file itself! No need for separate files in that case!

MichaelLeeHobbs:
Is this something coming in 0.7.0? I do not even see a way to pass in the mongo creds.

YourTechBud:
Unfortunately not. Currently you need to put the mongo credentials in the connection string itself (in the config). It will probably be in the next release.

What we are planning to do is, load the entire connection string from an environment variable.

MichaelLeeHobbs:
Can you give an example with creds?

YourTechBud:
Here's a link for reference:
https://docs.mongodb.com/manual/reference/connection-string/

Here's an example:
```yaml
conn: mongodb://username:secret-password@localhost:27017
```

MichaelLeeHobbs:
@YourTechBud Thanks, sorry I even asked. I had just woke up and wasn't thinking 😊

YourTechBud:
No problem. After implementing this feature the connection string can also be provided via an environment variable. Something like this would go in the config file:
```yaml
conn: $MY_MONGO_ENV
```

YourTechBud:
Fixed in #217. Will release in 0.9.0

PR #200, opened 2019-04-26
Enhancement: Connection parser – automatically parses connection string from info, can read passwords from either env or file

ollykel:
# Enhancement: Connection Parser
Hey there!

I had some time to contribute an enhancement this week. It's related to issue #150, which I raised last week about reading passwords from secure files. I went a step further and implemented an extension of the config package that automatically parses connection strings for each database using its components - user, password, hostname, etc. While doing this, it gives the option to read passwords from either an environment variable or a file by prefixing the password with either "ENV:" or "FILE:". The connection data is stored in a struct called ConnConfig (see config/conn.go), which is included in each CrudStub under the field name "Connection".

Below is the implementation of ConnConfig:
```go
// largely based on go-sql-driver's ConnConfig struct
// 	see: https://godoc.org/go-sql-driver/mysql#ConnConfig
// also based on lib/pq's connection string parameters
// 	see: https://godoc.org/github.com/lib/pq
type ConnConfig struct {
	User		string	`json:"user" yaml:"user"`
	// Auth provides source of password, either env or file
	Auth		string	`json:"auth" yaml:"auth"`
	DBName		string	`json:"dbName" yaml:"dbName"`
	// Protocol ex: "tcp"
	Protocol	string	`json:"protocol" yaml:"protocol"`
	Host		string	`json:"host" yaml:"host"`
	Port		string	`json:"port" yaml:"port"`
	SSLMode		sslMode	`json:"sslMode" yaml:"sslMode"`
	SSLCert		string	`json:"sslCert,omitempty" yaml:"sslCert,omitempty"`
	SSLKey		string	`json:"sslKey,omitempty" yaml:"sslKey,omitempty"`
	SSLRootCert	string	`json:"sslRootCert,omitempty" yaml:"sslRootCert,omitempty"`
	Params		map[string]string `json:"params,omitempty" yaml:"params,omitempty"`
}//-- end SQLConnConfig struct
```
Note: sslMode is based on lib/pq's sslMode param, which has four possible values: "require" , "disable", "verify-ca", and "verify-full". In databases with only true or false options for ssl, disable equals false and all other settings equal true.

The file config/parseconn.go contains functions to parse connection strings for all three of the supported databases from a ConnConfig struct.

The file config/generate.go has been modified to generate a ConnConfig struct from a survey and save it to the input struct during space-cloud init. This makes use of the method ConnConfig.FromSurvey(), which is located in the file config/conn.go.

To ensure backward compatibility, the original "Conn" field is still present in each CrudStub; if it is non-empty, space-cloud run defaults to using it instead of parsing a connection string.

## Specifying how to get a password
The ConnConfig struct, as shown above, contains a field 'Auth", which is where a user can specify how to get a password. There are three possible options for this: a plain string, a file, or an environment variable. These are specified by the prefixes "STRING", "FILE", and "ENV". Below is an example of each:
```yaml
connection:
...
    auth: STRING:my_password
    auth: FILE:/run/secrets/my_password.txt
    auth: ENV:MY_PASSWORD 
...
```

## Additional Changes:
The following changes were made to support this enhancement:
- Each of the structs in config/config.go has been modified to make use of "omitempty" field tag where appropriate. This prevents the yaml parser from writing the fields to a file if they are empty.
- The func writeConfig in config/generate.go has been rewritten to use a yaml encoder instead of an html template to write to the config file. Because of the previous change, the func does not write empty fields to the file, producing output identical to the output from the template.
- The template in config/template.go has been rewritten as a plain yaml string instead of a template. The func writeConfig uses yaml.Decoder.Decode() to generate a default Project struct from this template before adding the user input and writing to a file.


YourTechBud:
Hey buddy. Thats a lot of work i see here! Thanks a ton. I'll review this tomorrow or day after. To start with, can you squash all commits into one?

YourTechBud:
Hey @ollykel. I'm sorry i haven't been able to go through your code yet. Will clear this in a couple of weeks.

ollykel:
It's ok, it's a big commit haha. I've been busy this week too, I'll finish squashing the commits today.

Should I reopen the pull request on v0.8.0, btw? I didn't see that last time.

YourTechBud:
@ollykel yes. Change the base to v0.8.0.

YourTechBud:
Its too big a request. We should break this into smaller chunks. Closing it for now

PR #202, opened 2019-04-30
Fix/198 refactor faas module to functions module

ledex:
Changed faas to functions throughout the code-base. I am not sure if the plural is good here. 
@YourTechBud what do you think?


ledex:
Thats not what I wanted to do :D

PR #217, opened 2019-05-15
set conn string to environment var if provided (solves issue#150)

VedantParanjape:
Crud is a map of Crudstub, you have commented that key to the map is the database type, but i dont see any place where it is stored, i.e. taking input from yaml/json file.

### DIFF
--- a/config/load.go
+++ b/config/load.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"strings"
-
+    "os"
 	"gopkg.in/yaml.v2"
 )
 
@@ -25,6 +25,10 @@ func LoadConfigFromFile(path string) (*Project, error) {
 	}
 	if err != nil {
 		return nil, err
+	}
+
+	if conf.Modules.Crud["mongodb"].Conn[0] = "$"{
+		conf.Modules.Crud["mongodb"].Conn = os.Getenv(string.Trimprefix(conf.Modules.Crud["mongodb"].Conn), "$")
 	}
 
 	return conf, nil
--- a/config/load.go
+++ b/config/load.go
@@ -25,7 +25,7 @@ func loadEnvironmentVariable(p *Project) {
 			tempEnvVarC, presentC := os.LookupEnv(tempStringC)
 
 			if presentC {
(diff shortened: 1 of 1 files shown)


config/load.go
```diff
@@ -3,11 +3,34 @@ package config
...
+	"os"
 	"strings"
 
 	"gopkg.in/yaml.v2"
 )
 
+func loadEnvironmentVariable(p *Project) {
+	if strings.HasPrefix(p.Secret, "$") {
+		tempString := strings.TrimPrefix(p.Secret, "$")
+		tempEnvVar, doesItExist := os.LookupEnv(tempString)
```
YourTechBud: Let's rename doesItExist to `p` or `present`?

config/load.go
```diff
@@ -3,11 +3,34 @@ package config
...
+
+		if doesItExist {
+			p.Secret = tempEnvVar
+		}
+	}
+
+	for _, i := range p.Modules.Crud {
```
YourTechBud: Lets rename i to something like `c` (for crud) or `value`?

config/load.go
```diff
@@ -3,11 +3,34 @@ package config
...
+		if strings.HasPrefix(i.Conn, "$") {
+			tempStringC := strings.TrimPrefix(i.Conn, "$")
+			tempEnvVarC, doesItExistC := os.LookupEnv(tempStringC)
+
+			if doesItExistC {
```
YourTechBud: Again doesItExist can be `p` or `present`

config/load.go
VedantParanjape: okay, all done

config/load.go
```diff
@@ -3,11 +3,34 @@ package config
...
+		}
+	}
+
+	for _, value := range p.Modules.Crud {
+		if strings.HasPrefix(value.Conn, "$") {
+			tempStringC := strings.TrimPrefix(value.Conn, "$")
+			tempEnvVarC, presentC := os.LookupEnv(tempStringC)
+
+			if presentC {
+				value = tempEnvVarC
```
YourTechBud:
```suggestion
				value.Conn = tempEnvVarC
```

PR #230, opened 2019-05-21
embedded nats server

### LINKED ISSUE #211: Can space cloud have an internal pub sub network.
Currently Space Cloud relies on an external message broker for several modules like functions and realtime. This makes deployment a lil more difficult since the message broker needs to be set up as well.
### How can we solve it?
We can implement an internal pub sub network within Space Cloud. The `space-cloud` instances can be started with a `--seed-nodes` flag which can auto bootstrap a space cloud cluster.
>👍

YourTechBud:
fixes #211


nats.go
```diff
@@ -0,0 +1,27 @@
...
+		Port: 4248,
+	},
+}
+
+func (s *server) runNatsServer(opts *nats.Options) {
+	s.nats = nats.New(opts)
+	go s.nats.Start()
+	// Wait for accept loop(s) to be started
+	if !s.nats.ReadyForConnections(10 * time.Second) {
+		panic("Unable to start NATS Server in Go Routine")
```
Jayesh333: Can we throw error instead of panic?

YourTechBud: We need to force quit if this error occurs otherwise the other modules won't work. Can use a log.Fatal instead?

YourTechBud:
@Jayesh333 I have made the changes

PR #240, opened 2019-05-29
User Management GRPC

AliabbasMerchant:
Work in Progress, but @YourTechBud Please have a look.


proto/server.proto
```diff
@@ -99,6 +99,44 @@ message RealTimeResponse {
   string error = 4;
   repeated FeedData feedData = 5;
 }
+
+message UserManMeta {
```
YourTechBud: Just keep use the Meta object? `col` will be unused.

YourTechBud:
@AliabbasMerchant I dont think we have an edit profile for http. Lets keep the implementation like this:
- Id should always be required
- Having the email, name, role, pass is optional.
- Only those fields will be updated which are set.
- Lets make a `EditProfile` method in the `profile.go` file which takes an `EditProfileRequest`. `EditProfileRequest` will have an id and a map[string]string. Only those fields present in the map will be updated.
- The password will always be hashed and stored.

proto/server.proto
AliabbasMerchant: Ok

AliabbasMerchant:
Also, do we wanna use gRPC codes instead of HTTP codes?? Currently, the grpc_server is sending codes corres to http. gRPC has its own set of status codes. https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

YourTechBud:
@AliabbasMerchant lets stick to the http error codes to maintain consistency across all APIs.

modules/userman/operations.go
```diff
@@ -0,0 +1,361 @@
...
+	
+	result["status"] = http.StatusOK
+	result["result"] = res
+	return result
+}
+
+func (m *Module) EmailSignIn(ctx context.Context, dbType, project, email, password string) map[string]interface{} {
+	result := map[string]interface{}{"status": nil, "error": nil, "result": nil}
+	// Allow this feature only if the email sign in function is enabled
+	if !m.IsActive("email") {
```
YourTechBud: Let's use isEnabled here

AliabbasMerchant:
@YourTechBud  Done

modules/userman/operations.go
```diff
@@ -0,0 +1,361 @@
...
+
+	"golang.org/x/crypto/bcrypt"
+
+	uuid "github.com/satori/go.uuid"
+
+	"github.com/spaceuptech/space-cloud/model"
+	"github.com/spaceuptech/space-cloud/utils"
+)
+
+func (m *Module) Profile(ctx context.Context, token, dbType, project, id string) map[string]interface{} {
```
YourTechBud: Lets not return a map. Its makes the implementation unclear. Lets return status, result, err? Let result be a map.

modules/userman/operations.go
```diff
@@ -0,0 +1,361 @@
...
+	authObj, err := m.auth.IsAuthenticated(token, dbType, "users", utils.Update)
+	if err != nil {
+		result["status"] = http.StatusUnauthorized
+		result["error"] = err.Error()
+		return result
+	}
+
+	req := model.UpdateRequest{}
+	temp := map[string]interface{}{}
+	var id_string string
```
YourTechBud: we use camel case in go. Can we remove all snake case names?

modules/userman/operations.go
```diff
@@ -0,0 +1,361 @@
...
+
+	authObj, err := m.auth.IsAuthenticated(token, dbType, "users", utils.Update)
+	if err != nil {
+		result["status"] = http.StatusUnauthorized
+		result["error"] = err.Error()
+		return result
+	}
+
+	req := model.UpdateRequest{}
+	temp := map[string]interface{}{}
```
YourTechBud: Let's rename this to `find`

modules/userman/operations.go
```diff
@@ -0,0 +1,361 @@
...
+	var id_string string
+	if dbType == string(utils.Mongo) {
+		id_string = "_id"
+	} else {
+		id_string = "id"
+	}
+	temp[id_string] = id
+	req.Find = temp
+
+	temp = map[string]interface{}{}
```
YourTechBud: rename this to update?

modules/userman/operations.go
```diff
@@ -0,0 +1,361 @@
...
+	temp_set := map[string]interface{}{}
```
YourTechBud: rename this to set?

grpc_server.go
```diff
@@ -740,3 +741,52 @@ func (s *server) RealTime(stream pb.SpaceCloud_RealTimeServer) error {
...
+	return resp, err
+}
+
+func (s *server) SignUp(ctx context.Context, in *pb.SignUpRequest) (*pb.Response, error) {
+	result := s.user.EmailSignUp(ctx, in.Meta.DbType, in.Meta.Project, in.Email, in.Name, in.Password, in.Role)
+	resp, err := generate_response(result)
+	return resp, err
+}
+
+func generate_response(result map[string]interface{}) (*pb.Response, error) {
```
YourTechBud: Should be camel case.

modules/userman/operations.go
AliabbasMerchant: But then if there is no error, how do I return nil if I keep error as a string?

YourTechBud: Use type `error`

AliabbasMerchant:
@YourTechBud Please have a look

grpc_server.go
```diff
@@ -740,3 +741,50 @@ func (s *server) RealTime(stream pb.SpaceCloud_RealTimeServer) error {
...
+}
+
+func (s *server) EditProfile(ctx context.Context, in *pb.EditProfileRequest) (*pb.Response, error) {
+	status, err, result := s.user.EmailEditProfile(ctx, in.Meta.Token, in.Meta.DbType, in.Meta.Project, in.Id, in.Email, in.Name, in.Password)
+	resp := generateResponse(status, err, result)
+	return resp, nil
+}
+
+func (s *server) SignIn(ctx context.Context, in *pb.SignInRequest) (*pb.Response, error) {
+	status, err, result := s.user.EmailSignIn(ctx, in.Meta.DbType, in.Meta.Project, in.Email, in.Password)
```
YourTechBud: Err must me the last parameter according to golang conventions

grpc_server.go
```diff
@@ -740,3 +741,50 @@ func (s *server) RealTime(stream pb.SpaceCloud_RealTimeServer) error {
...
+	resp := generateResponse(status, err, result)
```
YourTechBud: Don't pass the error as a parameter. Handle it here itself. Maybe we don't need generateResponse?

modules/userman/http.go
```diff
@@ -0,0 +1,165 @@
...
+		id := vars["id"]
+
+		// Get the JWT token from header
+		tokens, ok := r.Header["Authorization"]
+		if !ok {
+			tokens = []string{""}
+		}
+		token := strings.TrimPrefix(tokens[0], "Bearer ")
+
+		status, err, result := m.Profile(ctx, token, dbType, project, id)
```
YourTechBud: Handle the error immediately after this line. Only have the condition, `err != nil`. We don't need an else if you return in the if itself

modules/userman/operations.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	delete(req, "pass")
+
+	// Create a new token Object
+	tokenObj := map[string]interface{}{
+		"email": email,
+		"role":  role,
+	}
+	tokenObj["id"] = id.String()
```
YourTechBud: This line can be shifted inside the map creation

grpc_server.go
AliabbasMerchant: Ok

grpc_server.go
AliabbasMerchant: Ok

AliabbasMerchant:
@YourTechBud Done. Please have a look.

PR #253, opened 2019-06-04
moved http functions to the `handlers` package

### LINKED ISSUE #251: Move http handler functions to utils/handlers package
Currently the http handler functions are located in their respective modules package. Some are even in http.go in the root. This makes it extremely scattered and difficult to navigate
### How can we solve it?
Lets move these handlers to the `utils/handlers` package
>👍

YourTechBud:
Fixes #251


utils/handlers/functions.go
```diff
@@ -12,8 +12,8 @@ import (
 	"github.com/spaceuptech/space-cloud/modules/functions"
 )
 
-// HandleRequest creates a Functions request endpoint
-func HandleRequest(functions *functions.Module, auth *auth.Module) http.HandlerFunc {
+// HandleFunctionCall creates a Functions request endpoint
+func HandleFunctionCall(functions *functions.Module, auth *auth.Module) http.HandlerFunc {
```
Inconnu08: In my opinion, should be called `HandleFunctions` or`HandleCustomFunctions`. Easier to understand. What do you think?

YourTechBud: This is the handler invoked when the `api.call()` method is called on the client. Technically it is handling a _function call_ event right? Hence the name HandleFunctionCall.
Makes sense?

Inconnu08: Ahh makes sense. Yeah, everything else seemed good.

PR #255, opened 2019-06-05
Add special type for the parsed token #248

### LINKED ISSUE #248: Make a special type for the parsed token
A lot of confusion is caused where a particular variable is an parsed token object or not.
### How can we solve it?
Make a new type called `TokenClaims` in the `utils` or `auth` package. This must the parameter and return type when ever the token claims are being passed around.
```golang
type TokenClaims map[string]interface{}
```
>👍

Inconnu08:
- Resolves #248
- Ran `GoImports`

### DIFF
--- a/modules/auth/auth.go
+++ b/modules/auth/auth.go
@@ -24,6 +24,9 @@ type Module struct {
 	project   string
 }
 
+// TokenClaims holds the JWT token claims
+type TokenClaims map[string]interface{}
+
 // Init creates a new instance of the auth object
 func Init(crud *crud.Module, functions *functions.Module) *Module {
 	return &Module{rules: make(config.Crud), crud: crud, functions: functions}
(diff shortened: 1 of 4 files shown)


YourTechBud:
@Inconnu08 lets resolve the merge conflicts? I'll merge this request right in after that.

Inconnu08:
@YourTechBud extremely sorry for the late response was afk.  Can you check if it's ok?

YourTechBud:
@Inconnu08 Looks like you got back the old code which we had changed. Hard reset back to your last commit, soft reset by another commit, stash your changes, then merge v0.9.0. Then apply your stashed changes.

Or maybe you should start over again. The last request we merged changed a lot of things in the auth module!

Inconnu08:
@YourTechBud woah! Might as well start over with the new changes. 😅

YourTechBud:
Okay. Closing this request

PR #257, opened 2019-06-06
 finished login page

ShreyaLaheri:


admin-ui/src/components/header/Header.jsx
```diff
@@ -0,0 +1,12 @@
+import React, { Component } from 'react'
+
+function Header(props) {
+  return (
+    <div>
+      <p className="welcome" style={{color: props.color, fontSize: props.fontSize}}>{props.name}</p>
```
Jayesh333: the className can be more general purpose like `heading` instead of `welcome`

admin-ui/src/pages/login/LoginForm.jsx
```diff
@@ -0,0 +1,63 @@
...
+  render() {
+
+    const { getFieldDecorator } = this.props.form;
+    return (
+      <div className="card">
+        <div className="content">
+          <img className="logo" src={logo} /><br />
+          <p className="about">If you've got an idea, let's space it up !</p>
+
+          <Form onSubmit={this.props.handleSubmit} className="login-form">
```
Jayesh333: Dont submit directly. Prevent the default form submit action and then call the props.handleSubmit

admin-ui/src/pages/login/LoginForm.jsx
```diff
@@ -0,0 +1,63 @@
...
+            <Form.Item className="input">
+              {getFieldDecorator('userName', {
+                rules: [{ required: true, message: 'Please input your username!' }],
```
Jayesh333: `u` in `username` here should be capital

PR #268, opened 2019-06-14
Realtime Fix for gRPC and SQL

AliabbasMerchant:
@YourTechBud  
SQL takes in an integer ID. I had to add the `AcceptableIdType` function to allow integers also.

### DIFF
--- a/grpc_server.go
+++ b/grpc_server.go
@@ -373,7 +373,7 @@ func (s *server) RealTime(stream pb.SpaceCloud_RealTimeServer) error {
 
 			// Subscribe to relaitme feed
 			feedData, err := s.realtime.Subscribe(ctx, clientID, s.auth, s.crud, data, func(feed *model.FeedData) {
-				client.Write(&model.Message{Type: utils.TypeRealtimeFeed, Data: feed})
+				client.Write(&model.Message{ID: req.ID, Type: utils.TypeRealtimeFeed, Data: feed})
 			})
 			if err != nil {
 				res := model.RealtimeResponse{Group: data.Group, ID: data.ID, Ack: false, Error: err.Error()}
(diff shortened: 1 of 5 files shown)


grpc_server.go
```diff
@@ -373,7 +373,7 @@ func (s *server) RealTime(stream pb.SpaceCloud_RealTimeServer) error {
 
 			// Subscribe to relaitme feed
 			feedData, err := s.realtime.Subscribe(ctx, clientID, s.auth, s.crud, data, func(feed *model.FeedData) {
-				client.Write(&model.Message{Type: utils.TypeRealtimeFeed, Data: feed})
+				client.Write(&model.Message{ID: req.ID, Type: utils.TypeRealtimeFeed, Data: feed})
```
YourTechBud: Why do we need to send the id here?

YourTechBud:
The implentation may pose this problem, the ids in the docs array will be of type string even when its an int in the database. Correct?

grpc_server.go
AliabbasMerchant: Just to be extra sure that we are responding to the correct client

AliabbasMerchant:
I think no. I got back an integer in the python client

grpc_server.go
YourTechBud: Okay. I don't think that is required. The client receiving the message is identified by the `client` object we are calling write on

YourTechBud:
Ah yes! You are right.

grpc_server.go
AliabbasMerchant: I am checking for the id at the client side. Serves as a double check. Like how we check during the Subscribe response. As it is, we have the id field empty. Lets use it. 😜

AliabbasMerchant:
I think yes.

grpc_server.go
YourTechBud: Lol. Okay

PR #270, opened 2019-06-14
Rules

SarveshKargutkar:


admin-ui/src/components/rules/rules.jsx
```diff
@@ -0,0 +1,82 @@
...
+							<div className="code-mirror">
+								<CodeMirror
+									value={values[index]}
+									options={{
+										mode: 'xml',
+										// theme: 'material',
+										autofocus: true,
+										lineNumbers: true
+									}}
+									onBeforeChange={(editor, data, value) => {
```
Jayesh333: You can directly provide this.handleChange function here to onBeforeChange since the params given to on before change and those expected by handleChange are same

admin-ui/src/components/rules/rules.jsx
```diff
@@ -0,0 +1,82 @@
...
+	handleChange(editor, data, value) {
+		console.log(value);
+		this.props.handleRuleChange(value);
+	}
+	handleClick(rule) {
+		this.setState({ selectedRule: rule });
+	}
+	render() {
+		var rules = Object.keys(this.props.rules);
+		const values = Object.values(this.props.rules);
```
Jayesh333: There is no need of making values and index here since you can directly access value of a particular rule by this.props.rules[this.state.selectedRule]

admin-ui/src/pages/database/rules/Rules.jsx
```diff
@@ -1,52 +1,74 @@
...
+					<div className="header-flex">
+						<Header name="Rules" color="#000" fontSize="22px" />
+						<Documentation url="https://spaceuptech.com/docs/database" />
+					</div>
+					<DbConfigure updateFormState={props.updateFormState} formState={props.formState} />
+					{/* <EmptyState graphics={rulesImg} desc="Guard your data with rules that define who has access to it and how it is structured." buttonText="Add a table" handleClick={props.handleClick} /> */}
+					<RulesTable
+						rules={props.rules}
+						handleRuleChange={props.handleRuleChange}
+						addText={props.addText}
```
Jayesh333: addText should not come from connect function. Pass the hard coded value to rules component as a prop from here only

admin-ui/src/pages/database/rules/Rules.jsx
```diff
@@ -1,52 +1,74 @@
...
+						handleAddTextClick={props.handleAddTextClick}
```
Jayesh333: It's should be props.handleAddTableClick

admin-ui/src/pages/database/rules/Rules.jsx
```diff
@@ -1,52 +1,74 @@
...
+		rules: {
+			Books: 'a',
+			Todos: 'b',
+			Users: 'c',
+			xyz: 'd',
+			sas: 'e',
+			qwqw: 'f',
+			rewwr: 'g'
+		},
+		addText: 'Add a table rule'
```
Jayesh333: Remove this from here

admin-ui/src/components/rules/rules.jsx
Jayesh333: Not resolved yet. You are still using values array

admin-ui/src/components/rules/Rules.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+		this.handleClick = this.handleClick.bind(this);
+	}
+
+	handleClick(rule) {
+		this.setState({ selectedRule: rule });
+	}
+	render() {
+		var rules = Object.keys(this.props.rules);
+		const index = rules.indexOf(this.state.selectedRule);
```
Jayesh333: Why is this index being made here. It's not getting used anywhere.

PR #277, opened 2019-06-16
configure done

SarveshKargutkar:


admin-ui/src/components/configure/FileStorageConfigure.css
```diff
@@ -0,0 +1,33 @@
+.fileStorage-configuration {
```
Jayesh333: Make just one configure.css file

PR #282, opened 2019-06-19
[WIP] File Management GRPC

AliabbasMerchant:
@YourTechBud Please have a look

### DIFF
--- a/proto/server.proto
+++ b/proto/server.proto
@@ -140,6 +140,37 @@ message SignUpRequest {
   Meta meta = 5;
 }
 
+message CreateFolderRequest {
+  string path = 1;
+  string name = 2;
+  Meta meta = 3;
+}
+message ListFilesRequest {
+  string path = 1;
+  Meta meta = 2;
+}
+message DeleteFileRequest {
+  string path = 1; // Full path, including the file name and extension
+  Meta meta = 2;
+}
+message DownloadFileRequest {
+  string path = 1; // Full path, including the file name and extension
+  Meta meta = 2;
+}
+message UploadFileRequest {
+  string path = 1;
+  string name = 2;
+  bytes payload = 3;
+  int64 size = 4; // Size of the whole file in bytes
+  string hash = 5; // Hash for the whole file
+  Meta meta = 6;
+}
+message FilePayload {
+  int64 size = 1; // Size of the whole file in bytes
+  string hash = 2; // Hash for the whole file
+  bytes payload = 3;
+}
+
 service SpaceCloud {
   rpc Create(CreateRequest) returns (Response) {}
   rpc Read(ReadRequest) returns (Response) {}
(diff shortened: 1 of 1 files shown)


YourTechBud:
change the base to v0.9.0

proto/server.proto
```diff
@@ -140,6 +140,37 @@ message SignUpRequest {
...
+message DownloadFileRequest {
+  string path = 1; // Full path, including the file name and extension
+  Meta meta = 2;
+}
+message UploadFileRequest {
+  string path = 1;
+  string name = 2;
+  bytes payload = 3;
+  int64 size = 4; // Size of the whole file in bytes
+  string hash = 5; // Hash for the whole file
```
YourTechBud: How are you planing to use the hash? If the file is too large it won't fit in the RAM. How will you hash then?

proto/server.proto
```diff
@@ -140,6 +140,37 @@ message SignUpRequest {
...
+}
+message DeleteFileRequest {
+  string path = 1; // Full path, including the file name and extension
+  Meta meta = 2;
+}
+message DownloadFileRequest {
+  string path = 1; // Full path, including the file name and extension
+  Meta meta = 2;
+}
+message UploadFileRequest {
```
YourTechBud: Which fields will be present in the subsequent requests?

proto/server.proto
```diff
@@ -140,6 +140,37 @@ message SignUpRequest {
...
+  string path = 1;
+  string name = 2;
+  bytes payload = 3;
+  int64 size = 4; // Size of the whole file in bytes
+  string hash = 5; // Hash for the whole file
+  Meta meta = 6;
+}
+message FilePayload {
```
YourTechBud: Which fields will be present in the subsequent payloads?

proto/server.proto
AliabbasMerchant: I was thinking that we store the file, and then check the hash.

proto/server.proto
AliabbasMerchant: path, name, payload, meta.
Do you want to keep a uuid also as a field?

proto/server.proto
AliabbasMerchant: Only the payload.
Do you want to keep a uuid also as a field?

proto/server.proto
YourTechBud: Let's add that in a future release? You can open an issue for it. Let's not do hash of now

proto/server.proto
YourTechBud: only send the payload

proto/server.proto
YourTechBud: nope. The payload is fine

proto/server.proto
AliabbasMerchant: Ok

proto/server.proto
AliabbasMerchant: Ok

PR #285, opened 2019-06-20
Added start command to admin ui

Jayesh333:


config/generate.go
```diff
@@ -56,17 +61,49 @@ func GenerateConfig() error {
...
+		}
+
+		// Ask for the admin password
+		err = survey.AskOne(&survey.Input{Message: "Mission Control (Password)", Default: "admin123"}, &i.AdminPass, survey.Required)
+		if err != nil {
+			return err
+		}
+
+		// Ask for the admin role
+		err = survey.AskOne(&survey.Input{Message: "Mission Control (Role)", Default: "super-admin"}, &i.AdminRole, survey.Required)
```
YourTechBud:
```suggestion
		err = survey.AskOne(&survey.Input{Message: "Mission Control (Role)", Default: "captain-cloud"}, &i.AdminRole, survey.Required)
```

config/generate.go
```diff
@@ -56,17 +61,49 @@ func GenerateConfig() error {
...
+		if err != nil {
+			return err
+		}
+
+		i.HomeDir = utils.UserHomeDir()
+		i.BuildVersion = utils.BuildVersion
+	}
+
+	if configFilePath == "none" {
+		configFilePath = "./" + i.ID + ".yaml"
```
YourTechBud: Use the path separator from the os package

YourTechBud: Let's avoid using `./` completely

config/generate.go
```diff
@@ -56,17 +61,49 @@ func GenerateConfig() error {
 		return err
 	}
 
-	return writeConfig(i)
+	if isMissionControlUIPresent {
```
YourTechBud: Should we ask admin details first?

config/template.go
```diff
@@ -51,8 +51,66 @@ modules:
...
+            rule: allow
+          read:
+            rule: allow
+          delete:
+            rule: allow
+  static:
+    enabled: true
+    routes:
+    - prefix: /mission-control
+      path: {{.HomeDir}}/space-cloud/mission-control-v{{.BuildVersion}}/build
```
YourTechBud:
```suggestion
      path: {{.HomeDir}}/.space-cloud/mission-control-v{{.BuildVersion}}/build
```

main.go
```diff
@@ -130,24 +124,115 @@ func actionRun(c *cli.Context) error {
...
+		err := config.GenerateConfig(configPath, true)
+		if err != nil {
+			return nil
+		}
+		temp, err := config.LoadConfigFromFile(configPath)
 		if err != nil {
 			return err
 		}
+
+		conf = temp
```
YourTechBud: We don't need this line

utils/file.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Create the file
+	out, err := os.Create(dest)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	// Write the body to file
+	_, err = io.Copy(out, resp.Body)
```
YourTechBud: Surround the file with a buffered reader?

config/generate.go
Jayesh333: Don't know. Let's keep it as it is only for now since we are asking project id in the beginning even in the init command.

main.go
Jayesh333: This line will be needed as conf is been used below in `s.LoadConfig(conf)`

YourTechBud: If line 172 is changed, this wouldn't be required

config/generate.go
YourTechBud: Cool.

main.go
Jayesh333: Cool got it!

utils/file.go
Jayesh333: The `io.Copy` function does not keeps the entire file in memory. It keeps writing as it receives in blocks of 32KB

main.go
Jayesh333: Done

config/generate.go
Jayesh333: Completed

utils/file.go
YourTechBud: Cool

PR #286, opened 2019-06-20
Admin login now sends project name on login

Jayesh333:


YourTechBud:
Push this change in #285.

PR #291, opened 2019-06-22
Fix FileMan Bugs #284

### LINKED ISSUE #284: FileMan Bugs
### Describe the bug
1. Does not work for Windows. The file separator does not match. Everywhere, we have mentioned "/", which is Unix Styled. We need to instead put `os.PathSeparator`
2. There is a small but dangerous bug. If we specify the filenames as `"/../"`, it allows us to insert into the parent folder, which is dangerous. We must use absolute paths to check.

>👍

Inconnu08:
- Fixes #284
- Addes corresponding test
- Fixes typos
- Bumps the version
- Ran Gofmt

### DIFF
--- a/modules/auth/file_test.go
+++ b/modules/auth/file_test.go
@@ -10,6 +10,7 @@ import (
 func TestGetFileRule(t *testing.T) {
 
 	var ps = string(os.PathSeparator)
+
 	fileRule := &config.FileRule{
 		Prefix: ps,
 		Rule:   map[string]*config.Rule{"rule": &config.Rule{Rule: "allow"}},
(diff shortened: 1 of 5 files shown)


modules/auth/file_test.go
```diff
@@ -25,10 +26,10 @@ func TestGetFileRule(t *testing.T) {
...
-		{testName: "Success", path: "/folder", module: &Module{fileRules: map[string]*config.FileRule{"create": fileRule1, "delete": fileRule1, "read": fileRule1}}},
-		{testName: "Success", path: "/folder/file", module: &Module{fileRules: map[string]*config.FileRule{"create": fileRule1, "delete": fileRule1, "read": fileRule1}}},
+		{testName: "Success", path: ps, module: &Module{fileRules: map[string]*config.FileRule{"create": fileRule, "delete": fileRule, "read": fileRule}}},
+		{testName: "Success", path: ps + "folder", module: &Module{fileRules: map[string]*config.FileRule{"create": fileRule, "delete": fileRule, "read": fileRule}}},
+		{testName: "Success", path: ps + "folder", module: &Module{fileRules: map[string]*config.FileRule{"create": fileRule1, "delete": fileRule1, "read": fileRule1}}},
+		{testName: "Success", path: ps + "folder" + ps + "file", module: &Module{fileRules: map[string]*config.FileRule{"create": fileRule1, "delete": fileRule1, "read": fileRule1}}},
 
 		//Error Test
 		{testName: "Fail", path: "/NewFolder/file", module: &Module{fileRules: map[string]*config.FileRule{"create": fileRule1, "delete": fileRule1, "read": fileRule1}}},
```
AliabbasMerchant: Please replace "/"s in the 35th line also with `ps`

AliabbasMerchant:
Also, @Inconnu08 please fix the `..` error.

> There is a small but dangerous bug. If we specify the filenames as "/../", it allows us to insert into the parent folder, which is dangerous. We must use absolute paths to check.

AliabbasMerchant:
@YourTechBud Looks fine to me. Please have a look

YourTechBud:
Perfect. @Inconnu08 love the fact that you correct the comments as well.

PR #295, opened 2019-06-26
Add pprof endpoints to space-cloud #293

### LINKED ISSUE #293: Include the profiler endpoints in the space-cloud binary
No way to profile cpu and ram usage for doing optimisation
### How can we solve it?
Include the profiler endpoints in the main binary. Have a `--profile` flag to explicitly enable it.
>👍

Inconnu08:
Resolves #293

### DIFF
--- a/main.go
+++ b/main.go
@@ -72,6 +72,11 @@ func main() {
 					Usage:  "Seed nodes to cluster with",
 					EnvVar: "SEEDS",
 				},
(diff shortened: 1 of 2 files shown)


YourTechBud:
@Inconnu08 did you check if it works?

Inconnu08:
@YourTechBud yup it works!

PR #296, opened 2019-06-27
Add a space cloud servers start up message

Inconnu08:
It's still not clear if Space Cloud servers are running or not (*at least that was the case for me*)

It should be more clear and consistent. And I think a timestamp when the servers started successfully should be shown. Hence this pr.

### DIFF
--- a/utils/server/server.go
+++ b/utils/server/server.go
@@ -69,8 +69,9 @@ func (s *Server) Start(port, grpcPort string) error {
 
 	handler := corsObj.Handler(s.router)
 
-	fmt.Println("Starting http Server on port " + port)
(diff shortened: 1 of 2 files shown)


utils/server/server.go
```diff
@@ -69,8 +69,9 @@ func (s *Server) Start(port, grpcPort string) error {
 
 	handler := corsObj.Handler(s.router)
 
-	fmt.Println("Starting http Server on port " + port)
+	fmt.Println("Starting HTTP Server on port: " + port)
 
+	log.Printf("Space Cloud is running on the specified ports...")
```
YourTechBud:
```suggestion
	log.Printf("Space Cloud is running on the specified ports :D")
```

PR #302, opened 2019-06-30
FileMan Auth is now an array

### LINKED ISSUE #298: Lets use an array instead of a map for the file man security rules
This is important since the alignment of a map can change during json parsing. We need to preserve the order of these rules.

AliabbasMerchant:
Fixes #298

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -96,7 +96,7 @@ type FileStore struct {
 	Enabled   bool                 `json:"enabled" yaml:"enabled"`
 	StoreType string               `json:"storeType" yaml:"storeType"`
 	Conn      string               `json:"conn" yaml:"conn"`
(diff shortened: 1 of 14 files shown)


Jayesh333:
Will have to change the template and examples as well @AliabbasMerchant

AliabbasMerchant:
@Jayesh333 Right. I forgot that. Have made the changes in the latest commit

AliabbasMerchant:
@YourTechBud @Jayesh333 I have added docs for FileStore.  Please do not squash the commits while merging

YourTechBud:
@AliabbasMerchant is the indentation for the rules array correct?

AliabbasMerchant:
Yes. It is correct. And it is working properly. I checked

PR #318, opened 2019-07-03
Minor bug fixes

### LINKED ISSUE #299: `utils.exists` method in security rules doesn't work for `auth` object
### Describe the bug
Whenever we use the `utils.exists` method in security rules to validate whether a field is present in the `auth` object, the rule fails giving an 403 with the error `incorrect` field type
### Expected behaviour
The rule should be validated if the condition (i.e, the required field was present in `auth` obj) was correct.
### Actual behaviour 
Instead of considering the request to be an authorised one, its considering the request to be unauthorised
### Steps to reproduce
Try writing a security rule with `utils.exists` 

### Your environment
- Space Cloud version: 0.9.0
- OS: linux

>👍

### LINKED ISSUE #301: The result sent by call of functions module is inconsistent with the crud module
### Describe the bug
The functions module returns the result directly instead of sending it in a map with the `result` field, as in CRUD module.
### Expected behaviour
The functions module must return the result in the same fashion as crud module.

>👍

AliabbasMerchant:
Fixes #299  
Fixes #301

### DIFF
--- a/modules/auth/match_types.go
+++ b/modules/auth/match_types.go
@@ -95,13 +95,16 @@ func matchBool(rule *config.Rule, args map[string]interface{}) error {
 	case "==":
 		if f1 == f2 {
 			return nil
+		} else {
+			return ErrIncorrectMatch
 		}
 
 	case "!=":
 		if f1 != f2 {
 			return nil
+		} else {
+			return ErrIncorrectMatch
 		}
 	}
-
 	return ErrIncorrectRuleFieldType
 }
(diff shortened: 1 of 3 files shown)


YourTechBud:
@AliabbasMerchant are you sure this solved #299. I think you'll have to type case the auth object as a `map[string]interface{}` before passing it to the load value.

AliabbasMerchant:
I checked it. It is working properly, and giving the proper error

YourTechBud:
Did you try applying utils.exist to the auth object?

YourTechBud:
@AliabbasMerchant ?

AliabbasMerchant:
@YourTechBud 
I checked it.


config:
```
rule: match
eval: ==
type: bool
f1: utils.exists(args.auth.id)
f2: true
```  
When the token contains id in the claim, it gives no error.  
When the token does not contain the id, it gives `Auth: The two fields do not match`

PR #319, opened 2019-07-03
Added Changes for LiveQuery Options

AliabbasMerchant:
@YourTechBud Done.  
This replaces the previous PR. (Which I have closed)


YourTechBud:
@AliabbasMerchant is this complete? You still seem to be firing the get request internally event when initial is supposed to be neglected. Can we avoid that query?

YourTechBud:
@AliabbasMerchant we can't use `AddLiveQuery` directly. That bypasses the entire security module.

AliabbasMerchant:
@YourTechBud Please have a look. I have tested it and is working.

modules/realtime/realtime.go
```diff
@@ -54,6 +54,10 @@ type pendingRequest struct {
 	time time.Time
 }
 
+type LiveQueryOptions struct {
```
YourTechBud: Define this in the `model` package

model/realtime.go
```diff
@@ -20,6 +20,7 @@ type RealtimeRequest struct {
 	Type    string                 `json:"type"`  // Can either be subscribe or unsubscribe
 	ID      string                 `json:"id"`    // id is the query id
 	Where   map[string]interface{} `json:"where"`
+	Options map[string]interface{} `json:"options"`
```
YourTechBud: Keep this of type `LiveQueryOptions`

modules/realtime/operations.go
```diff
@@ -21,6 +23,14 @@ func (m *Module) Subscribe(ctx context.Context, clientID string, auth *auth.Modu
 		return nil, err
 	}
 
+	options := new(LiveQueryOptions)
+	mapstructure.Decode(data.Options, data)
+
+	if options.SkipInitial {
+		m.AddLiveQuery(data.ID, data.Project, data.Group, clientID, data.Where, sendFeed)
+		return nil, nil
```
YourTechBud: Lets return an empty array of feed data?

utils/client/grpc_realtime.go
```diff
@@ -90,6 +90,12 @@ func (c *GRPCRealtimeClient) Read(cb DataCallback) {
 			return
 		}
 		data["Where"] = temp
+		err = json.Unmarshal(in.Options, &temp)
```
YourTechBud: Don't forget to clear temp before this step. Re initialise temp.

YourTechBud: Temp should be of type LiveQueryOptions i guess

utils/server/grpc.go
```diff
@@ -408,8 +409,13 @@ func (s *Server) RealTime(stream pb.SpaceCloud_RealTimeServer) error {
 				return true
 			}
 
+			options := new(realtime.LiveQueryOptions)
```
YourTechBud: Lets remove these steps complete. Lets send the empty feed data array as is. On the client we can ignore the message if feed data array is empty

utils/server/websocket.go
```diff
@@ -54,8 +55,13 @@ func (s *Server) handleWebsocket() http.HandlerFunc {
 					return true
 				}
 
+				options := new(realtime.LiveQueryOptions)
```
YourTechBud: This part is not required

AliabbasMerchant:
@YourTechBud Perfect! Your requested changes made the logic quite easy to understand.  
I have made the changes.  
This the requested changes do not break any of the clients and do not change the client implementation.

model/realtime.go
```diff
@@ -20,6 +20,7 @@ type RealtimeRequest struct {
 	Type    string                 `json:"type"`  // Can either be subscribe or unsubscribe
 	ID      string                 `json:"id"`    // id is the query id
 	Where   map[string]interface{} `json:"where"`
+	Options LiveQueryOptions       `json:"options"`
```
YourTechBud: LiveQueryOptions must be a pointer. Prefix it with *

model/realtime.go
```diff
@@ -37,3 +38,7 @@ type Message struct {
 	Data interface{} `json:"data"`
 	ID   string      `json:"id"` // the request id
 }
+
+type LiveQueryOptions struct {
```
YourTechBud: Add the export comment

model/realtime.go
```diff
@@ -37,3 +38,7 @@ type Message struct {
...
+	SkipInitial bool `json:"skipInitial"`
```
YourTechBud: Wasn't this supposed to be change only. Lets use a consistent nomenclature across

modules/realtime/operations.go
```diff
@@ -21,6 +23,14 @@ func (m *Module) Subscribe(ctx context.Context, clientID string, auth *auth.Modu
 		return nil, err
 	}
 
+	options := new(model.LiveQueryOptions)
+	mapstructure.Decode(data.Options, options)
```
YourTechBud: Why do you need this step?

model/realtime.go
AliabbasMerchant: I think no. They are 2 diff things. The server has nothing to do with `changesOnly`. That is handled entirely on the client. The server just needs to know whether to send the initial or not. (Internally, in the clients, I am setting the values of both to be the same)

modules/realtime/operations.go
AliabbasMerchant: Oh. Right

model/realtime.go
YourTechBud: Cool. Works for me.

PR #348, opened 2019-07-24
Added GCP Storage

AliabbasMerchant:
@YourTechBud Done
I have tested also. It is working


modules/auth/handle_file.go
```diff
@@ -39,21 +39,25 @@ func (m *Module) IsFileOpAuthorised(project, token, path string, op utils.FileOp
...
 func (m *Module) getFileRule(path string) (map[string]interface{}, *config.FileRule, error) {
 	pathParams := make(map[string]interface{})
+	ps := "/"
+	if m.fileStoreType == string(utils.Local) {
+		ps = string(os.PathSeparator)
+	}
 
 	// Check if its a valid absolute path
-	if strings.Contains(path, string(os.PathSeparator) + "..") {
+	if strings.Contains(path, ps + "..") {
```
YourTechBud: Let's check if it contains a `..` no need for the prefixed /

modules/filestore/amazons3/create.go
```diff
@@ -12,30 +11,31 @@ import (
 )
 
 // CreateFile creates a file in S3
-func (a *AmazonS3) CreateFile(ctx context.Context, project string, req *model.CreateFileRequest, file io.Reader) error {
+func (a *AmazonS3) CreateFile(project string, req *model.CreateFileRequest, file io.Reader) error {
+	path := strings.Trim(req.Path, "/")
+	name := strings.Trim(req.Name, "/")
+	p := strings.Trim(path + "/" + name, "/")
```
YourTechBud: You sure this will work?

AliabbasMerchant: Yes. It works

modules/auth/handle_file.go
AliabbasMerchant: Done

PR #353, opened 2019-07-25
FixedIssue346

### LINKED ISSUE #346: Decouple bucket name from project id in fileStore
There is no way to decouple the bucket name of `amazon-s3` from the project id
### How can we solve it?
We can solve this by providing a bucket name in the config file or either in the request directly. If we are mentioning the bucket name in the request itself, then there needs to be some form of mechanism to mention which bucket names are allowed
>👍

AliabbasMerchant:
@YourTechBud Done

Fixes #346

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -143,6 +143,7 @@ type FileStore struct {
 	StoreType string      `json:"storeType" yaml:"storeType"`
 	Conn      string      `json:"conn" yaml:"conn"`
 	Endpoint  string      `json:"endpoint" yaml:"endpoint"`
+	Bucket    string      `json:"bucket" yaml:"bucket"`
 	Rules     []*FileRule `json:"rules" yaml:"rules"`
 }
 
--- a/docs/manual/file-storage/config.md
+++ b/docs/manual/file-storage/config.md
@@ -24,8 +24,9 @@ modules:
 Inside `fileStore` you have to speicify the following fields:
 - **enable:** Setting this field to `true` enables the file storage module.
 - **storeType:** Choose the storage technology to back the file storage module. The possible options are `local`, `amazon-s3` and `gcp-storage`.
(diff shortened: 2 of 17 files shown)


modules/crud/sql/read.go
```diff
@@ -69,18 +69,30 @@ func (s *SQL) Read(ctx context.Context, project, col string, req *model.ReadRequ
...
+	case utils.Count:
+		query = query.Select(goqu.COUNT("*"))
+	case utils.Distinct:
+		distinct := req.Options.Distinct
+		if distinct == nil {
+			return nil, utils.ErrInvalidParams
+		}
+		query = query.SelectDistinct(*distinct)
+	case utils.One, utils.All:
+		query = query.Select(selArray...)
```
YourTechBud: What will happen if select array is empty? Do we get the results?

modules/crud/sql/read.go
```diff
@@ -101,6 +113,27 @@ func (s *SQL) Read(ctx context.Context, project, col string, req *model.ReadRequ
...
+			return nil, err
+		}
+
+		switch s.GetDBType() {
+		case utils.MySQL, utils.Postgres:
+			mysqlTypeCheck(rowTypes, mapping)
+		}
+
+		for _, v := range mapping {
+			return v, nil
```
YourTechBud: What are we exactly returning here? The count number? I think we should return an object with key count. Whats the mongo api returning?

modules/crud/sql/read.go
AliabbasMerchant: Yes

modules/crud/sql/read.go
AliabbasMerchant: Yes, we are returning a number. The mongo api also returns a count number.

PR #354, opened 2019-07-28
Added Support for More Options in SQL

AliabbasMerchant:


modules/crud/sql/read.go
```diff
@@ -69,18 +69,30 @@ func (s *SQL) Read(ctx context.Context, project, col string, req *model.ReadRequ
...
+	case utils.Count:
+		query = query.Select(goqu.COUNT("*"))
+	case utils.Distinct:
+		distinct := req.Options.Distinct
+		if distinct == nil {
+			return nil, utils.ErrInvalidParams
+		}
+		query = query.SelectDistinct(*distinct)
+	case utils.One, utils.All:
+		query = query.Select(selArray...)
```
YourTechBud: What if the sel array is empty,?

modules/crud/sql/read.go
```diff
@@ -101,6 +113,27 @@ func (s *SQL) Read(ctx context.Context, project, col string, req *model.ReadRequ
...
+			return nil, err
+		}
+
+		switch s.GetDBType() {
+		case utils.MySQL, utils.Postgres:
+			mysqlTypeCheck(rowTypes, mapping)
+		}
+
+		for _, v := range mapping {
+			return v, nil
```
YourTechBud: Shouldn't we return an object here? What does mongo return currently? We should make sql seem similar

modules/crud/sql/read.go
```diff
@@ -128,11 +161,6 @@ func (s *SQL) Read(ctx context.Context, project, col string, req *model.ReadRequ
 				return nil, err
 			}
 
-			switch s.GetDBType() {
-			case utils.MySQL, utils.Postgres:
-				rowTypes, _ = rows.ColumnTypes()
```
YourTechBud: Why did you remove this?

modules/crud/sql/update.go
```diff
@@ -1,60 +1,226 @@
...
+		}
+	case utils.Upsert:
+		for k := range req.Update {
+			switch k {
+			case "$set", "$inc", "$mul", "$max", "$min", "$currentDate":
+				err := s.upsert(ctx, project, col, req, k)
+				if err != nil {
+					return err
+				}
+			default: // (case "$push", "$unset", "$rename")
```
YourTechBud: Should we silently ignore this or throw an error? @Jayesh333 what do you think?

modules/crud/sql/update.go
```diff
@@ -1,60 +1,226 @@
...
+	if req == nil {
+		return utils.ErrInvalidParams
 	}
-	return s.doExecContext(ctx, sqlString, args)
+	if req.Update == nil {
+		return utils.ErrInvalidParams
+	}
+	switch req.Operation {
+	case utils.All:
+		for k := range req.Update {
```
YourTechBud: We shouldn't have a for loop here. `generateUpdateQuery` should internal have this for loop in my opinion.

modules/crud/sql/update.go
```diff
@@ -1,60 +1,226 @@
...
 	sqlString = strings.Replace(sqlString, "\"", "", -1)
+	switch op {
+	case "$set":
+	case "$inc":
+		for k, v := range m {
+			val, err := numToString(v)
+			if err != nil {
+				return "", nil, err
+			}
+			sqlString = strings.Replace(sqlString, k+"="+val, k+"="+k+"+"+val, -1)
```
YourTechBud: This is prone to injection attacks.
Why aren't we using a transaction? Such operations aren't truly atomic in all cases. We'll explicitly set the consistency level to make sure we have a record level lock so no other query can simultaneously perform a read or write.

modules/crud/sql/update.go
```diff
@@ -1,60 +1,226 @@
...
+	case "$currentDate":
+		for k, v := range m {
+			val, ok := v.(string)
+			if !ok {
+				return "", nil, utils.ErrInvalidParams
+			}
+			if val == "timestamp" {
+				sqlString = strings.Replace(sqlString, k+"='"+val+"'", k+"=CURRENT_TIMESTAMP()", -1)
+			} else {
+				sqlString = strings.Replace(sqlString, k+"='"+val+"'", k+"=CURRENT_DATE()", -1)
```
YourTechBud: Do we have `CURRENT_DATE()` and `CURRENT_TIMESTAMP()` exist for both, postgres and MySQL?

modules/crud/sql/update.go
```diff
@@ -1,60 +1,226 @@
...
+	}
+	if rowsAffected == 0 {
+		if op == "$currentDate" {
+			return utils.ErrInvalidParams
+		}
+		sqlQuery, args, err := s.generateCreateQuery(ctx, project, col, &model.CreateRequest{Document: req.Update[op], Operation: utils.One})
+		if err != nil {
+			return err
+		}
+		err = doExecContext(ctx, sqlQuery, args, tx)
```
YourTechBud: You are firing one create query for each update op. Seems to be wrong and it won't have goot performance

YourTechBud:
@AliabbasMerchant did you make those changes?

AliabbasMerchant:
@YourTechBud Done!

modules/crud/sql/update.go
```diff
@@ -1,60 +1,236 @@
...
+			}
+			for op := range req.Update {
+				if op == "$currentDate" {
+					return utils.ErrInvalidParams
+				}
+				m, ok := req.Update[op].(map[string]interface{})
+				if !ok {
+					return utils.ErrInvalidParams
+				}
+				for k, v := range m { // k -> $max, $min, $inc, $mul
```
YourTechBud: Need to correct the comment here

modules/crud/sql/update.go
```diff
@@ -1,60 +1,236 @@
...
+		defer rows.Close()
+		
+		if !rows.Next() {
+			// not found. So, insert
+			doc := make(map[string]interface{})
+			for k, v := range req.Find {
+				doc[k] = v
+			}
+			for op := range req.Update {
+				if op == "$currentDate" {
```
YourTechBud: We need to be able to support `$currentDate` in upserts too

PR #381, opened 2019-08-07
handled disabled and connection errors properly

Jayesh333:


modules/crud/sql/sql.go
```diff
@@ -77,6 +66,38 @@ func (s *SQL) GetDBType() utils.DBType {
...
+	return nil
+}
+
+func (s *SQL) connect() error {
+	timeOut := 5 * time.Second
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeOut)
+	defer cancel()
+
+	sql, err := sqlx.Open(s.dbType, s.connection)
```
YourTechBud: The `dbType` will have values like `sql-postgres` & `sql-mysql` I guess. You can't pass those directly I guess

PR #384, opened 2019-08-11
Added Pubsub module

AliabbasMerchant:
@YourTechBud Please don't review now. The mapping part is left. I will do that, and inform you


modules/pubsub/broker.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+natsWildcard string = ">"
+
+type subscription interface {
+	Unsubscribe() error
+}
+
+type pubsubBroker interface {
+	Publish(subject string, msg *model.PubsubMsg) error
```
Jayesh333: Isn't subject getting duplicated over here?. It's also there in PubsubMsg.

modules/pubsub/broker.go
```diff
@@ -0,0 +1,101 @@
...
+	QueueSubscribe(subject, group string, ch chan *model.PubsubMsg) (*subscription, error)
```
Jayesh333: There should also be a normal subscribe right?

modules/pubsub/broker.go
```diff
@@ -0,0 +1,101 @@
...
+	subs *subscription
+	ch   chan *model.PubsubMsg
+}
+
+func (m *Module) Publish(subject string, data *model.PubsubMsg) error {
+	return m.connection.Publish(subject, data)
+}
+
+func (m *Module) Subscribe(clientId, subject string, cb model.PubsubCallback) (error) {
+	return m.QueueSubscribe(clientId, subject, clientId, cb)
```
Jayesh333: The arguments seem to be incorrect. You are passing 4 paremeters whereas it requires only 3. Have you build the code? And shouldn't the subject here be also single leading?

model/pubsub.go
```diff
@@ -0,0 +1,9 @@
+package model
+
+// PubsubMsg is the message recevied by the subscriber
+type PubsubMsg struct {
+	Data    []byte
+	Subject string
+}
+
+type PubsubCallback func(subject string, msg *PubsubMsg) ()
```
YourTechBud: Do we need to pass the subject here when its already present in the struct? I guess we should remove it from the struct

modules/pubsub/broker.go
YourTechBud: Lets use `queue` instead of `group`

modules/pubsub/broker.go
```diff
@@ -0,0 +1,101 @@
...
+	m.storage.Range(func(subject, v interface{}) bool {
+		m.Unsubscribe(clientId, subject)
+		return true
+	})
+	return nil
+}
+
+//Map subject -> clientId -> pubsubSubscription
+func (m *Module) storeSubs(subject, clientId string, subs *pubsubSubscription) {
+	client, ok := m.storage.Load(subject)
```
YourTechBud: Use LoadOrStore here

modules/pubsub/broker.go
```diff
@@ -0,0 +1,101 @@
...
+	if ok {
+		cli := client.(sync.Map)
+		cli.Store(client, subs)
+		return;
+	}
+	var m sync.Map
+	m.Store(client, subs)
+	m.storage.Store(subject, m)
```
YourTechBud: Seems like you are overwriting all clients here. Instead you must check if a client with that id already exists using LoadOrStore. If it does, throw an error.

modules/pubsub/broker.go
```diff
@@ -0,0 +1,101 @@
...
+
+//Map subject -> clientId -> pubsubSubscription
+func (m *Module) storeSubs(subject, clientId string, subs *pubsubSubscription) {
+	client, ok := m.storage.Load(subject)
+	if ok {
+		cli := client.(sync.Map)
+		cli.Store(client, subs)
+		return;
+	}
+	var m sync.Map
```
YourTechBud: Rename this. Took me 5 minutes to understand that its a map and not the module

modules/pubsub/broker.go
```diff
@@ -0,0 +1,101 @@
...
+func (m *Module) QueueSubscribe(clientId, subject, group string, cb model.PubsubCallback) (error) {
+	ch := make(chan *model.PubsubMsg, 10)
+	subject = utils.SingleLeadingTrailing(subject, "/")
+	subs, err := m.connection.QueueSubscribe(subject + natsWildcard, group, ch)
+	if err != nil {
+		return err
+	}
+	go func() {
+		for msg := range ch {
+        	cb(msg.Subject, msg)
```
YourTechBud: Use the subject from the parameter

modules/pubsub/broker.go
```diff
@@ -0,0 +1,101 @@
...
+	if ok {
+		subj := subj.(sync.Map)
+		_subs, ok := subj.Load(clientId)
+		if ok {
+			subs := _subs.(pubsubSubscription)
+			subj.Delete(clientId)
+			if mapLen(subj) == 0 {
+				m.storage.Delete(subject)
+			}
+			subj.Range(func (key, value interface{}))
```
Jayesh333: There are errors over here

modules/pubsub/broker.go
```diff
@@ -0,0 +1,101 @@
...
+	go func() {
+		for msg := range ch {
+        	cb(msg.Subject, msg)
+		}
+	}()
+	m.storeSubs(subject, clientId, &pubsubSubscription{subs, ch})
+	return nil
+}
+
+func (m *Module) Unsubscribe(clientId, subject string) err {
```
YourTechBud: You need to fix the logic for this

modules/pubsub/broker.go
AliabbasMerchant: Sorry

modules/pubsub/broker.go
AliabbasMerchant: If user subscibes to "/a/"(the subject in the parameter), and some one publishes to "/a/b/", we send the subject as "/a/b/"(the subject in the msg), right?

YourTechBud: You are correct. My bad

modules/pubsub/broker.go
YourTechBud: @AliabbasMerchant i suggest you keep a nornal subscribe as well. Internally call QueueSubscribe.

AliabbasMerchant: This is not exposed to the outside world. This is just for internal use.

YourTechBud: I understand. But let that logic be inside.

YourTechBud: Actually its alright

modules/pubsub/broker.go
YourTechBud: This is correct @Jayesh333

Jayesh333: Yess

modules/pubsub/broker.go
```diff
@@ -0,0 +1,105 @@
...
+	receiveChannel := make(chan *model.PubsubMsg)
+	subject = utils.SingleLeadingTrailing(subject, "/")
+	subs, err := m.connection.QueueSubscribe(subject, queue, receiveChannel)
+	if err != nil {
+		return err
+	}
+	subscription := pubsubSubscription{subs, receiveChannel}
+	err = m.storeSubs(subject, clientId, &subscription)
+	if err != nil {
+		subscription.unsubscribeFunc()
```
YourTechBud: It would be better is this function internally closes the channel.

YourTechBud: Lets just call the function, `unsubscribe`

modules/pubsub/broker.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (m *Module) Unsubscribe(clientId, subject string) error {
+	subject = utils.SingleLeadingTrailing(subject, "/")
+	subjM, ok := m.storage.Load(subject)
+	if ok {
+		subjectMap := subjM.(sync.Map)
+		subs, ok := subjectMap.Load(clientId)
+		if ok {
+			subscription := subs.(pubsubSubscription)
```
YourTechBud: You are strong this object as a pointer but retrieving it the other way

modules/pubsub/nats/nats.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &nats{c}, nil
+}
+
+func (n *nats) Publish(subject string, msg *model.PubsubMsg) error {
+	sendChannel := make(chan *model.PubsubMsg)
+	// defer sendChannel.Close()
+	err := n.conn.BindSendChan(subject, sendChannel)
```
YourTechBud: Why can't we simply publish. We don't really need this here.

modules/pubsub/nats/nats.go
```diff
@@ -0,0 +1,41 @@
...
+	return nil
+}
+
+func (n *nats) QueueSubscribe(subject, queue string, ch chan *model.PubsubMsg) (func() (error) , error) {
+	natsWildcard := ">"
+	subs, err := n.conn.BindRecvQueueChan(subject + natsWildcard, queue, ch)
+	if err != nil {
+		return nil, err
+	}
+	return subs.Unsubscribe, nil
```
YourTechBud: Maybe you should return pubsubUnsubscribe from here

modules/pubsub/pubsub.go
```diff
@@ -0,0 +1,65 @@
...
+
+	if pubsub == nil || !pubsub.Enabled {
+		m.enabled = false
+		return nil
+	}
+
+	switch pubsub.Broker {
+	case utils.Nats:
+		m.broker = Pubsub.Broker
+		if m.connection == nil {
```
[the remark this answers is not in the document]
YourTechBud: Lets not do that. Even in the other modules we have refrained from doing that. It will disturb the exisiting subscriptions.

modules/pubsub/nats/nats.go
AliabbasMerchant: Without bindingChannels, we can only send a byte array. And this is the ideal way to use nats.
"EncodedConn are the preferred way to interface with NATS"

modules/pubsub/broker.go
AliabbasMerchant: > It would be better is this function internally closes the channel.
That would be difficult. I will have to wrap the nats subscription's unsubscribe function. Currently I am sending the nats' unsubscribe function directly

modules/pubsub/broker.go
AliabbasMerchant: Oh. Right

modules/pubsub/nats/nats.go
AliabbasMerchant: Okay

AliabbasMerchant:
@YourTechBud @Jayesh333 This is a very big commit (force push). Its compiling well. Still left to test it out.

modules/pubsub/nats/nats.go
YourTechBud: I think marshal that struct and use publish directly. It fits our use case better. This can be a potential memory leak.

modules/pubsub/operations.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// m.RLock()
+	// defer m.RUnlock()
+	
+	var err error = nil
+	status := 200
+
+	m.storage.Range(func(subject, v interface{}) bool {
+		s, e := m.Unsubscribe(clientID, subject.(string))
+		if e != nil {
```
YourTechBud: Ignore the error here. Try and unsubscribe for each subject

modules/pubsub/broker.go
YourTechBud: So do that!

modules/pubsub/nats/nats.go
YourTechBud: Not yet resolved

modules/auth/handle_pubusb.go
```diff
@@ -0,0 +1,98 @@
...
+func (m *Module) IsSubscribeAuthorised(project, token, subject, queue string, args map[string]interface{}) error {
+	return m.isPubsubAuthorised(project, token, subject, queue, utils.Subscribe, args)
+}
+
+// IsPublishAuthorised checks if the caller is authorized to make the publish request
+func (m *Module) IsPublishAuthorised(project, token, subject string, args map[string]interface{}) error {
+	return m.isPubsubAuthorised(project, token, subject, "", utils.Publish, args)
+}
+
+func (m *Module) getPubsubRule(subject string) (map[string]interface{}, *config.PubsubRule, error) {
```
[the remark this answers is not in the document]
AliabbasMerchant: Yes

modules/pubsub/nats/nats.go
AliabbasMerchant: The interface wont work correctly then. Instead of a generic channel, I will have to use the nats.Msg channel. This wont work properly with other pubsub providers.

AliabbasMerchant: I think the current thing will not lead to a memory leak, but still. let me research a bit more.

modules/pubsub/operations.go
AliabbasMerchant: Ok

modules/pubsub/broker.go
AliabbasMerchant: Cool

modules/pubsub/nats/nats.go
AliabbasMerchant: Done in the latest commit

modules/pubsub/broker.go
AliabbasMerchant: Done in the latest commit.

modules/pubsub/nats/nats.go
YourTechBud: Can we simply use this
https://godoc.org/github.com/nats-io/go-nats#EncodedConn.Publish

AliabbasMerchant: Cool. I'll do that

AliabbasMerchant: Just noticed, this does not have the queue option.

AliabbasMerchant: What if I close the channel in defer? We wont have a memory leak then, right?

YourTechBud: Do you need to specify the queue while publishing? Lets stick to `EncodedConn.Publish()`.

AliabbasMerchant: Cool

AliabbasMerchant:
@YourTechBud @Jayesh333 The whole thing is finally done.  
I have tested it also

PR #387, opened 2019-08-17
added insert operation in graphql

atharva29:


utils/graphql/delete.go
```diff
@@ -0,0 +1,58 @@
...
+
+	deleteRequest.Find, err = extractWhereClause(field.Arguments, store)
+	if err != nil {
+		return nil, err
+	}
+
+	return &deleteRequest, nil
+}
+
+func extractWhereClause(args []*ast.Argument, store m) (m, error) {
```
YourTechBud: Why are you re declaring this function?

utils/graphql/update.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return m{"status": status}, graph.crud.Update(context.TODO(), dbType, graph.project, col, req)
+}
+
+func generateUpdateRequest(field *ast.Field, store m) (*model.UpdateRequest, error) {
+	var err error
+	var updateRequest model.UpdateRequest
+
+	//---	 upsert identification is wrong
+	if strings.HasPrefix(field.Name.Value, "upsert") {
```
YourTechBud: This if else condition isn't needed! Instead make a `extractUpdateOperstion` function which returns the operation. The operation is available as the `op` field in the arguments.

utils/graphql/update.go
```diff
@@ -0,0 +1,80 @@
...
+
+	updateRequest.Update, err = extractUpdateArgs(field.Arguments, store)
+	if err != nil {
+		return nil, err
+	}
+
+	return &updateRequest, nil
+}
+
+func extractWhereClause(args []*ast.Argument, store m) (m, error) {
```
YourTechBud: Dont redeclare this function again

utils/graphql/update.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+	}
+	return m{}, nil
+}
+
+func extractUpdateArgs(args []*ast.Argument, store m) (m, error) {
+	var t map[string]interface{}
+	for _, v := range args {
+		switch v.Name.Value {
+		case "set", "inc", "mul", "max", "min", "currentTimestamp", "currentDate":
```
YourTechBud: `push`, `rename`, `remove` are missing

atharva29: ok

utils/graphql/update.go
atharva29: ok

PR #388, opened 2019-08-18
Pubsub Changes for JS

AliabbasMerchant:


YourTechBud:
You should take the subject and the payload in the body. Lets avoid the `.` all together.

model/pubsub.go
```diff
@@ -24,3 +24,8 @@ type PubsubMsgResponse struct {
 	Error  string `json:"error"`
 	Msg    []byte `json:"msg"`
 }
+
+type PubsubPublishRequest struct {
+	Subject string `json:"subject"`
+	Data    string `json:"data"`
```
YourTechBud: Shouldn't this be an interface?

utils/handlers/pubsub.go
```diff
@@ -36,7 +36,7 @@ func HandlePublishCall(pubsub *pubsub.Module, auth *auth.Module) http.HandlerFun
 		}
 		token := strings.TrimPrefix(tokens[0], "Bearer ")
 
-		status, err := pubsub.Publish(project, token, subject, b)
+		status, err := pubsub.Publish(project, token, req.Subject, []byte(req.Data))
```
YourTechBud: Dont take a byte array as a parameter, take an interface. You should marshal inside.

AliabbasMerchant:
@YourTechBud Done

utils/handlers/pubsub.go
YourTechBud: Parse an interface directly here. No need to marshal again inside the Publish() function

AliabbasMerchant:
@YourTechBud Done

PR #390, opened 2019-08-21
added endpoint to graphql

atharva29:


model/graphql.go
```diff
@@ -0,0 +1,7 @@
+package model
+
+type GraphQLRequest struct {
+	Query         string                 `json:"query"`
+	OperationName string                 `json:"op"`
```
YourTechBud: Shouldn't this be `operationName`? Confirm with the GraphQL spec

utils/graphql/helpers.go
```diff
@@ -112,6 +112,12 @@ func parseValue(value ast.Value, store m) (interface{}, error) {
 
 		return val, nil
 
+	case kinds.Variable:
+		t := value.(*ast.Variable)
+		//	e := store["vars"]
+		//////// ----
+		return utils.LoadValue(t.Name.Value, store)
```
YourTechBud: This should be:
```golang
return utils.LoadValue("vars." + t.Name.Value, store)
```

utils/handlers/graphql.go
```diff
@@ -0,0 +1,36 @@
...
+
+		// Get the path parameters
+		_ = getRequestMetaData(r)
+
+		// Load the request from the body
+		req := model.GraphQLRequest{}
+		json.NewDecoder(r.Body).Decode(&req)
+		defer r.Body.Close()
+
+		_, err := graphql.ExecGraphQLQuery(&req)
```
YourTechBud: We need to return the output to the client as well

utils/handlers/graphql.go
```diff
@@ -0,0 +1,36 @@
...
+
+		// Create a context of execution
+		_, cancel := context.WithTimeout(r.Context(), 5*time.Second)
+		defer cancel()
+
+		// Get the path parameters
+		_ = getRequestMetaData(r)
+
+		// Load the request from the body
+		req := model.GraphQLRequest{}
```
YourTechBud: For a GET request, the query will be as a query parameter. Add that as well

utils/handlers/graphql.go
```diff
@@ -0,0 +1,36 @@
...
+// HandleCrudCreate creates the create operation endpoint
+func HandleGraphQLRequest(graphql *graphql.Module) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+
+		// Create a context of execution
+		_, cancel := context.WithTimeout(r.Context(), 5*time.Second)
+		defer cancel()
+
+		// Get the path parameters
+		_ = getRequestMetaData(r)
```
YourTechBud: Use the project returned by this method to match with the project stored in the GraphQL module. Also, pass the token in the ExecGraphQLQuery method. We need to pass the token to wherever the auth module is used.

model/graphql.go
atharva29: Is it compulsory to have first letter in caps ?

utils/graphql/helpers.go
atharva29: ok

model/graphql.go
YourTechBud: Im talking about the json part. Let the variable name be as is

atharva29: ok

utils/handlers/graphql.go
```diff
@@ -18,19 +19,32 @@ func HandleGraphQLRequest(graphql *graphql.Module) http.HandlerFunc {
 		_, cancel := context.WithTimeout(r.Context(), 5*time.Second)
 		defer cancel()
 
-		// Get the path parameters
-		_ = getRequestMetaData(r)
+		vars := mux.Vars(r)
+		projectID := vars["project"]
+		pid := graphql.GetProjectID()
+		if projectID == pid {
```
YourTechBud: Instead of checking for equality, check for inequality and return if its unequal

utils/handlers/graphql.go
```diff
@@ -18,19 +19,32 @@ func HandleGraphQLRequest(graphql *graphql.Module) http.HandlerFunc {
...
 
-		// Load the request from the body
-		req := model.GraphQLRequest{}
-		json.NewDecoder(r.Body).Decode(&req)
-		defer r.Body.Close()
+			// Get the path parameters
+			meta := getRequestMetaData(r)
```
YourTechBud: Why don't you get the token here directly

PR #391, opened 2019-08-25
Schema graphql

sharadregoti:


modules/crud/schema.go
```diff
@@ -0,0 +1,120 @@
...
+			getFieldType(ve.Type, &fieldTypeStuct)
+			for _, w := range ve.Directives {
+				fieldTypeStuct.DirectiveType[w.Name.Value] = nil // directive name
+				argValue := map[string]string{}
+				for _, x := range w.Arguments {
+
+					val, _ := (utils.ParseGraphqlValue(x.Value, nil))
+					argValue[x.Name.Value] = val.(string) // direvtive field name & value name
+				}
+				fieldTypeStuct.DirectiveType[w.Name.Value] = argValue
```
YourTechBud: We should call `DirectiveType` as `properties`. Lets not use `maps` for this. Let's create a property struct.

modules/crud/schema.go
```diff
@@ -0,0 +1,146 @@
...
+			if colName != strings.Title(collectionName) {
+				continue
+			}
+			fieldTypeStuct := SchemaFieldType{
+				Directive: DirectiveProperties{
+					Value: DirectiveArgs{},
+				},
+			}
+
+			for _, w := range ve.Directives {
```
YourTechBud: Instead of the loop, check if the length of the `ve.Directives` array is more than zero. If yes, directly pick the 0th directive and work on that

modules/crud/schema.go
```diff
@@ -0,0 +1,146 @@
...
+			if fieldType.(*ast.List).Type.GetKind() == kinds.Named {
+				getFieldType(fieldType.(*ast.List).Type, fieldTypeStuct, doc)
+			} else {
+				getFieldType(fieldType.(*ast.List).Type.(*ast.NonNull).Type, fieldTypeStuct, doc)
+			}
+		}
+	case kinds.Named:
+		{
+			myType := fieldType.(*ast.Named).Name.Value
+			fmt.Println("Type:", myType)
```
YourTechBud: Remove all logs and prints which aren't required

PR #393, opened 2019-08-28
added schema validation function

sharadregoti:


modules/crud/schema.go
```diff
@@ -3,17 +3,21 @@ package crud
...
 	"github.com/spaceuptech/space-cloud/config"
+	"github.com/spaceuptech/space-cloud/model"
 	"github.com/spaceuptech/space-cloud/utils"
 )
 
+// TODO: check graphql types
+
 // ParseSchema Initializes Schema field in Module struct
-func (m *Module) ParseSchema(crud config.Crud) (SchemaType, error) {
+func (m *Module) ParseSchema(crud config.Crud) error {
```
YourTechBud: Need not export it since it will be called internally only

modules/crud/schema.go
```diff
@@ -132,7 +137,116 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
 	return nil
 }
 
+func (m *Module) schemaValidator(dbType, col string, fields map[string]interface{}) error {
+
+	collectionFields := m.schema[dbType][col]
```
YourTechBud: Check if the dbtype and col exists in the map. If it doesn't return nil.

modules/crud/schema.go
```diff
@@ -132,7 +137,116 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+	for fieldKey, fieldValue := range collectionFields {
+		// TODO: check if internal elements of list is required for this we need to change schmea parser
```
YourTechBud: Is something still pending?

modules/crud/schema.go
```diff
@@ -132,7 +137,116 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
 	return nil
 }
 
+func (m *Module) schemaValidator(dbType, col string, fields map[string]interface{}) error {
```
YourTechBud: Don't call it fields, call it `doc`

modules/crud/schema.go
```diff
@@ -132,7 +137,116 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+			}
+		}
+
+		// check type
+		fieldType, err := m.checkType(value, fieldValue.Kind, dbType, fieldValue.TableJoin)
+		if err != nil {
+			return err
+		}
+
+		if fieldType != fieldValue.Kind {
```
YourTechBud: Shouldn't this be inside `checkType()`

modules/crud/schema.go
```diff
@@ -132,7 +137,116 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+	}
+
+	return nil
+}
+
 // ValidateSchema checks data type
-func (m *Module) ValidateSchema() {
+func (m *Module) ValidateSchema(dbType, col string, req *model.CreateRequest) error {
+
+	v := make([]map[string]interface{}, 1)
```
YourTechBud: Let the length be zero

YourTechBud: And call it `docs` instead of v

modules/crud/schema.go
```diff
@@ -132,7 +137,116 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+
+	switch t := req.Document.(type) {
+	case []map[string]interface{}:
+		v = t
+	case map[string]interface{}:
+		v = append(v, t)
+	}
+
+	for _, fields := range v {
```
YourTechBud: Lets call it `doc` instead of fields

modules/crud/schema.go
```diff
@@ -132,7 +137,116 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+		err := m.schemaValidator(dbType, col, fields)
```
YourTechBud: Wrap the `if` in this line itself

modules/crud/schema.go
```diff
@@ -132,7 +137,116 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+}
+
+func (m *Module) checkType(value interface{}, kind, dbType, col string) (string, error) {
+
+	switch v := value.(type) {
+	case int:
+		// TODO: int64
+		switch kind {
+		case TypeDateTime:
+			unitTimeInRFC3339 := time.Unix(int64(v), 0).Format(time.RFC3339)
```
YourTechBud: Do we need to do that `.Format`

YourTechBud: Weren't you supposed to overwrite the field with a date object?

modules/crud/schema.go
```diff
@@ -132,7 +137,116 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+			if unitTimeInRFC3339 == "" {
+				return "", errors.New("Integer Wrong Date-Time Format")
+			}
+			return TypeDateTime, nil
+		case TypeID:
+			return TypeID, nil
+		}
+		return TypeInteger, nil
```
YourTechBud: What if the kind was string? You shouldn't return the kind. instead return an error on default

modules/crud/schema.go
```diff
@@ -132,7 +137,116 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+		case TypeDateTime:
+			_, err := time.Parse(time.RFC3339, v)
+			if err != nil {
+				return "", errors.New("String Wrong Date-Time Format")
+			}
+			return TypeDateTime, nil
+		case TypeID:
+			return TypeID, nil
+		}
+		return TypeString, nil
```
YourTechBud: What if the kind was not string

modules/crud/schema.go
```diff
@@ -132,7 +137,116 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+
+	case bool:
+		return TypeBoolean, nil
+
+	case map[string]interface{}:
+		err := m.schemaValidator(dbType, col, v)
+		if err != nil {
+			return "", err
+		}
+		return TypeJoin, nil
```
YourTechBud: What if its an embedded document

modules/crud/schema.go
sharadregoti: if the schema is defined as [String!]!
we only store information about [] list, if it is nullable or not
we do not store information about String if it is nullable or not

To implement the above we need to make some changes in parseSchema Function

modules/crud/schema.go
sharadregoti: ok, ill do that

modules/crud/schema.go
sharadregoti: > 
> 
> Do we need to do that `.Format`

func main() {
    theValue := 1405544146
    unixTimeUTC:=time.Unix(theValue, 0) //gives unix time stamp in utc 

    unitTimeInRFC3339 :=unixTimeUTC.Format(time.RFC3339) // converts utc time to RFC3339 format

    fmt.Println("unix time stamp in UTC :--->",unixTimeUTC)
    fmt.Println("unix time stamp in unitTimeInRFC3339 format :->",unitTimeInRFC3339)
}

This is something that u sent that' why I added it
and graphql spec. uses rfc3339

sharadregoti: > 
> 
> Weren't you supposed to overwrite the field with a date object?

which field u are talking about.
The actual field of type dateTime

modules/crud/schema.go
sharadregoti: kind is of type **String**

modules/crud/schema.go
sharadregoti: kind is always string it' nested switch case on type kind which is string

modules/crud/schema.go
sharadregoti: do u mean nested objects

sharadregoti: it will work for nested objects

modules/crud/schema.go
YourTechBud: Do we need that information?

modules/crud/schema.go
YourTechBud: So any number is mostly a valid date time object. We don't need to format it as a string at all.

In the docs object. We need to update the field corresponding to the date time kind right?

modules/crud/schema.go
YourTechBud: No no. I meant what if the value was `String`? You'll still return TypeInteger

modules/crud/schema.go
YourTechBud: If the value of `kind` was `Integer`, you are returning `TypeString`. Thats wrong

modules/crud/schema.go
YourTechBud: But you are returning the wrong type in this case. Even if this works, someone reading it might think there is a problem here.

modules/crud/schema.go
YourTechBud: Don't format things here. Return a type `time.Time` directly. No need to reassign value

modules/crud/schema.go
```diff
@@ -132,7 +137,126 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+		}
+
+	case string:
+		switch fieldValue.Kind {
+		case TypeDateTime:
+			unitTimeInRFC3339, err := time.Parse(time.RFC3339, v)
+			if err != nil {
+				return nil, errors.New("String Wrong Date-Time Format")
+			}
+			value = unitTimeInRFC3339
```
YourTechBud: No need to reassign value. Return the Time object directly

modules/crud/schema.go
```diff
@@ -132,7 +137,126 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+		}
+	case map[string]interface{}:
+		err := m.schemaValidator(fieldValue.NestedObject, v)
+		if err != nil {
+			return nil, err
+		}
+		return value, nil
+
+	case []interface{}:
+		arr := []interface{}{}
```
YourTechBud: Why don't you make an array with a fixed length `len(v)`

YourTechBud: Don't forget to set the array elements instead of appending

modules/crud/schema.go
```diff
@@ -132,7 +141,133 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+	collection, ok := m.schema[dbType]
+	if !ok {
+		return errors.New("No db was found named " + dbType)
+	}
+	collectionFields, ok := collection[col]
+	if !ok {
+		return errors.New("No collection or table was found named " + col)
+	}
+
+	for index, fields := range v {
```
YourTechBud: Rename `fields` to `doc`

modules/crud/schema.go
```diff
@@ -132,7 +141,133 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+}
+
+func (m *Module) checkType(value interface{}, fieldValue *SchemaFieldType) (interface{}, error) {
+
+	switch v := value.(type) {
+	case int:
+		// TODO: int64
+		switch fieldValue.Kind {
+		case TypeDateTime:
+			unitTimeInRFC3339 := time.Unix(int64(v), 0).Format(time.RFC3339)
```
YourTechBud:
```suggestion
			dateValue := time.Unix(int64(v), 0)
```

modules/crud/schema.go
```diff
@@ -3,38 +3,47 @@ package crud
...
 // ParseSchema Initializes Schema field in Module struct
-func (m *Module) ParseSchema(crud config.Crud) (SchemaType, error) {
+func (m *Module) parseSchema(crud config.Crud) error {
+
 	schema := make(SchemaType, len(crud))
 	for dbName, v := range crud {
 		collection := SchemaCollection{}
 		for collectionName, v := range v.Collections {
+			if v.Schema == "" {
+				return errors.New("Schema is empty")
```
YourTechBud:
```suggestion
				continue
```

modules/crud/schema.go
```diff
@@ -132,7 +141,133 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+		}
+		mutatedDoc[fieldKey] = val
+	}
+
+	return mutatedDoc, nil
+}
+
 // ValidateSchema checks data type
-func (m *Module) ValidateSchema() {
+func (m *Module) ValidateSchema(dbType, col string, req *model.CreateRequest) error {
```
YourTechBud: Need not export this

modules/crud/schema.go
```diff
@@ -132,7 +141,133 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
 	return nil
 }
 
+func (m *Module) schemaValidator(collectionFields SchemaField, doc map[string]interface{}) (map[string]interface{}, error) {
+	mutatedDoc := map[string]interface{}{}
+
+	for fieldKey, fieldValue := range collectionFields {
```
YourTechBud: return `doc` if `len(collectionFields) == 0`

modules/crud/schema.go
```diff
@@ -132,7 +141,133 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+		}
+	case bool:
+		switch fieldValue.Kind {
+		case TypeBoolean:
+			return value, nil
+		default:
+			return nil, errors.New("Bool wrong type wanted " + fieldValue.Kind + " got Bool")
+		}
+	case map[string]interface{}:
+		newDoc, err := m.schemaValidator(fieldValue.NestedObject, v)
```
YourTechBud:
```suggestion
		return m.schemaValidator(fieldValue.NestedObject, v)
```

modules/crud/schema.go
```diff
@@ -132,7 +141,133 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+
 // ValidateSchema checks data type
-func (m *Module) ValidateSchema() {
+func (m *Module) ValidateSchema(dbType, col string, req *model.CreateRequest) error {
+
+	if m.schema == nil {
+		return errors.New("Schema not initialized")
+	}
+
+	v := make([]map[string]interface{}, 0)
```
YourTechBud:
```suggestion
	v := make([]interface{}, 0)
```

modules/crud/schema.go
```diff
@@ -132,7 +141,133 @@ func getFieldType(fieldType ast.Type, fieldTypeStuct *SchemaFieldType, doc *ast.
...
+
+	switch t := req.Document.(type) {
+	case []map[string]interface{}:
```
YourTechBud:
```suggestion
	case []interface{}:
```

modules/crud/schema.go
YourTechBud: You'll need to type assert `doc` to `map[string]interface{}`

sharadregoti: > 
> 
> You'll need to type assert `doc` to `map[string]interface{}`

why

modules/crud/schema.go
sharadregoti: > 

.format returns a string so that I can compare if it's empty

so now should i check if dateValue is zero using isZero() function & return error

YourTechBud: No need to compare. Directly return it

modules/crud/schema.go
YourTechBud: Because `v` is `[]interface{}` nor `[]map[string]interface{}`

sharadregoti: > 
> 
> You'll need to type assert `doc` to `map[string]interface{}`

should i handle panic error if assertion is not possible

YourTechBud: nope

PR #402, opened 2019-09-09
added postgres inpection

sharadregoti:


modules/auth/schema/inspection.go
```diff
@@ -29,10 +31,15 @@ func (s *Schema) SchemaInspection(ctx context.Context, dbType, project, col stri
...
 
 		// field type
-		if err := inspectionCheckFieldType(value.FieldType, &fieldDetails); err != nil {
-			return "", err
+		if utils.DBType(dbType) == utils.Postgres {
+			if err := inspectionPostgresCheckFieldType(value.FieldType, &fieldDetails); err != nil {
+				return "", err
+			}
+		} else {
+			if err := inspectionCheckFieldType(value.FieldType, &fieldDetails); err != nil {
```
YourTechBud: make this `inspectionMySQLCheckFieldType`

modules/crud/crud.go
```diff
@@ -28,7 +28,7 @@ type Crud interface {
 	Delete(ctx context.Context, project, col string, req *model.DeleteRequest) error
 	Aggregate(ctx context.Context, project, col string, req *model.AggregateRequest) (interface{}, error)
 	Batch(ctx context.Context, project string, req *model.BatchRequest) error
-	DescribeTable(ctc context.Context, project, col string) ([]utils.FieldType, []utils.ForeignKeysType, error)
+	DescribeTable(ctc context.Context, project, col, dbType string) ([]utils.FieldType, []utils.ForeignKeysType, error)
```
YourTechBud: `dbType` should be before `col`

modules/crud/sql/describe.go
```diff
@@ -2,27 +2,56 @@ package sql
...
+	foreignKeys, err := s.getForeignKeyDetails(ctx, project, col, dbType)
 	if err != nil {
 		return nil, nil, err
 	}
 	return fields, foreignKeys, nil
 }
 
-func (s *SQL) getDescribeDetails(ctx context.Context, col string) ([]utils.FieldType, error) {
-	rows, err := s.client.Queryx("DESCRIBE " + col)
+func (s *SQL) getDescribeDetails(ctx context.Context, col, dbType string) ([]utils.FieldType, error) {
```
YourTechBud: `dbType` should be before `col`

modules/crud/sql/describe.go
```diff
@@ -2,27 +2,56 @@ package sql
...
+		END AS "Null"
+	FROM pg_attribute f  
+		JOIN pg_class c ON c.oid = f.attrelid  
+		JOIN pg_type t ON t.oid = f.atttypid  
+		LEFT JOIN pg_attrdef d ON d.adrelid = c.oid AND d.adnum = f.attnum  
+		LEFT JOIN pg_namespace n ON n.oid = c.relnamespace  
+		LEFT JOIN pg_constraint p ON p.conrelid = c.oid AND f.attnum = ANY (p.conkey)  
+		LEFT JOIN pg_class AS g ON p.confrelid = g.oid  
+	WHERE c.relkind = 'r'::char    
+		AND c.relname = '` + col + `'  -- Replace with table name
```
YourTechBud: take col as `:col`. And remove the table name

modules/crud/sql/describe.go
```diff
@@ -41,9 +70,30 @@ func (s *SQL) getDescribeDetails(ctx context.Context, col string) ([]utils.Field
...
+		ccu.column_name AS "REFERENCED_COLUMN_NAME"
+	FROM 
+		information_schema.table_constraints AS tc 
+		JOIN information_schema.key_column_usage AS kcu
+		  ON tc.constraint_name = kcu.constraint_name
+		  AND tc.table_schema = kcu.table_schema
+		JOIN information_schema.constraint_column_usage AS ccu
+		  ON ccu.constraint_name = tc.constraint_name
+		  AND ccu.table_schema = tc.table_schema
+	WHERE tc.constraint_type = 'FOREIGN KEY' AND tc.table_name='` + col + `'
```
YourTechBud: Take `col` as `:col`. Pass argument to the `Queryx` function below

modules/crud/sql/describe.go
sharadregoti: I didn't get u, what's u want to say

modules/crud/sql/describe.go
sharadregoti: same here

modules/crud/sql/describe.go
sharadregoti: it is not possible here, i have tried
because of sql describe

modules/crud/sql/describe.go
sharadregoti: done

sharadregoti:
> Make sure that the MySQL queries to describe the table do not have string manipulation.

statement describe for MySQL is not accepting positional $1 like params and named params
whereas Postgres accepts both

as query can be of MySQL or Postgres

I cannot call queryx function for only Postgres

modules/crud/sql/describe.go
YourTechBud: Do this. Make an array of string. In case of mysql it will be empty, in case of Postgres it would have the database and col. Then spread this array while querying.

YourTechBud: MySQL statements will have `?` while Postgres will have `$1`

modules/crud/sql/describe.go
```diff
@@ -2,27 +2,56 @@ package sql
...
 }
 
-func (s *SQL) getDescribeDetails(ctx context.Context, col string) ([]utils.FieldType, error) {
-	rows, err := s.client.Queryx("DESCRIBE " + col)
+func (s *SQL) getDescribeDetails(ctx context.Context, project, dbType, col string) ([]utils.FieldType, error) {
+	queryString := ""
+	if utils.DBType(dbType) == utils.MySQL {
+		queryString = `DESCRIBE ` + project + "." + col
+	} else {
+		queryString = `SELECT
```
YourTechBud: Which database is this searching for the table in?

modules/crud/sql/describe.go
```diff
@@ -41,9 +70,29 @@ func (s *SQL) getDescribeDetails(ctx context.Context, col string) ([]utils.Field
...
 
-func (s *SQL) getForeignKeyDetails(ctx context.Context, project, col string) ([]utils.ForeignKeysType, error) {
-	fmt.Println("Sql", project, col)
-	rows, err := s.client.Queryx("select TABLE_NAME, COLUMN_NAME, CONSTRAINT_NAME, REFERENCED_TABLE_NAME, REFERENCED_COLUMN_NAME FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE WHERE REFERENCED_TABLE_SCHEMA = '" + project + "' and TABLE_NAME = '" + col + "'")
+func (s *SQL) getForeignKeyDetails(ctx context.Context, project, dbType, col string) ([]utils.ForeignKeysType, error) {
+	queryString := ""
+	if utils.DBType(dbType) == utils.MySQL {
+		queryString = "select TABLE_NAME, COLUMN_NAME, CONSTRAINT_NAME, REFERENCED_TABLE_NAME, REFERENCED_COLUMN_NAME FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE WHERE REFERENCED_TABLE_SCHEMA = :project and TABLE_NAME = :col"
+	} else {
+		queryString = `SELECT
```
YourTechBud: Which database is this going to search the table in

modules/crud/sql/describe.go
YourTechBud: This works?

sharadregoti: yes

modules/crud/sql/describe.go
sharadregoti: query needs to be changes for that, will solve tmr with ur help

modules/crud/sql/describe.go
sharadregoti: same here

PR #406, opened 2019-09-11
schema creation

sharadregoti:
some query changes still need to be made


YourTechBud:
Let me know when its complete. I'll review it then

modules/auth/schema/creation.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return err
+	}
+	realSchema := parsedSchema[dbType]
+
+	batchedQueries := []string{}
+
+	for realColName, realColValue := range realSchema {
+		// check if table exist in current schema
+		currentColValue := currentSchema[realColName]
```
YourTechBud: Should you check if the key is present in `currentSchema`?

modules/auth/schema/creation.go
```diff
@@ -0,0 +1,210 @@
...
+				batchedQueries = append(batchedQueries, queries...)
+				if err != nil {
+					return err
+				}
+
+			}
+		}
+	}
+	for currentColName, currentColValue := range currentSchema {
+		realColValue := realSchema[currentColName]
```
YourTechBud: Should you check if the `realSchema` has the `currentColName` ?

modules/crud/operations.go
```diff
@@ -124,3 +124,20 @@ func (m *Module) DescribeTable(ctx context.Context, dbType, project, col string)
 
 	return crud.DescribeTable(ctx, project, dbType, col)
 }
+
+// RawBatch performs a db operaionf for schema creation
```
YourTechBud: fix the comment

modules/auth/schema/creation.go
sharadregoti: the key here is the table name
what if the current schema doesn't have the table specified by realColName
as in schema creation, we have not written any query for creating a table
so that' why I wanted to make sure that the table exists before altering the columns of that table
& return an error if not present

YourTechBud: So this is correct?

YourTechBud:
Add the table creation logic as well

PR #411, opened 2019-09-18
Graphql websocket

sharadregoti:


YourTechBud:
@sharadregoti you need to resolve this merge conflicts. Let me know once you are done

utils/server/websocket.go
```diff
@@ -152,3 +162,144 @@ func (s *Server) handleWebsocket() http.HandlerFunc {
...
+func (s *Server) handleGraphqlSocket(adminMan *admin.Manager) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		project := mux.Vars(r)["project"]
+		socket, err := upgrader.Upgrade(w, r, nil)
+		if err != nil {
+			log.Println("upgrade:", err)
+			return
+		}
+		defer socket.Close()
+
```
YourTechBud: Make a unique string named `clientID` here. Use UUID.

utils/server/websocket.go
```diff
@@ -152,3 +162,144 @@ func (s *Server) handleWebsocket() http.HandlerFunc {
...
+				data.Token = token
+				data.DBType = v.Directives[0].Name.Value
+				data.Project = project
+				data.Group = v.Name.Value
+				data.Type = m.Type
+				data.ID = m.ID
+				graphqlIDMapper.Store(m.ID, data.Group)
+
+				// Subscribe to realtime feed
+				feedData, err := s.realtime.Subscribe(ctx, m.ID, s.auth, s.crud, data, func(feed *model.FeedData) {
```
YourTechBud:
```suggestion
				feedData, err := s.realtime.Subscribe(ctx, clientID, s.auth, s.crud, data, func(feed *model.FeedData) {
```

utils/server/websocket.go
```diff
@@ -152,3 +162,144 @@ func (s *Server) handleWebsocket() http.HandlerFunc {
...
+				// For realtime subscribe event
+				group, ok := graphqlIDMapper.Load(m.ID)
+				if !ok {
+					return
+				}
+				data := new(model.RealtimeRequest)
+				data.ID = m.ID
+				data.Group = group.(string)
+
+				s.realtime.Unsubscribe(m.ID, data)
```
YourTechBud:
```suggestion
				s.realtime.Unsubscribe(clientID, data)
```

utils/server/websocket.go
YourTechBud: also call `defer s.realtime.RemoveClient(clientID)`

utils/constants.go
```diff
@@ -245,6 +245,29 @@ const (
 	Subscribe PubsubType = "subscribe"
 )
 
+const (
+	// GqlConnectionKeepAlive send every 20 second to client over websocket
+	GqlConnectionKeepAlive = "ka" // Server -> Client
```
YourTechBud: All these are not matching the types specified in the protocol. Everything should be in capital.

utils/server/websocket.go
```diff
@@ -152,3 +162,144 @@ func (s *Server) handleWebsocket() http.HandlerFunc {
...
+		ticker := time.NewTicker(20 * time.Second)
+		defer ticker.Stop()
+
+		for {
+
+			m := graphqlMessage{}
+			if err := socket.ReadJSON(m); err != nil {
+				return
+			}
+			select {
```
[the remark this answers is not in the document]
sharadregoti: ok, but how do i pass message id,(m.id)
i can pass the status only GqlConnectionKeepAlive

utils/server/websocket.go
```diff
@@ -152,3 +162,144 @@ func (s *Server) handleWebsocket() http.HandlerFunc {
...
+				source := source.NewSource(&source.Source{
+					Body: []byte(m.Payload.Query),
+				})
+				// parse the source
+				doc, err := parser.Parse(parser.ParseParams{Source: source})
+				if err != nil {
+					channel <- (&graphqlMessage{ID: m.ID, Payload: payloadObject{Status: utils.GqlConnectionError, Error: err.Error()}})
+					return
+				}
+				v := doc.Definitions[0].(*ast.OperationDefinition).SelectionSet.Selections[0].(*ast.Field)
```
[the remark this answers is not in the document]
sharadregoti: for every type assertions or the last one?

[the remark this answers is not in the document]
YourTechBud: For everyone

utils/server/websocket.go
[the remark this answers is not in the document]
YourTechBud: In that case you'll have to create a ticker for each start. And make sure you close them all when websocket disconnects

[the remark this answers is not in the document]
YourTechBud: Or maybe, you already have a map for all the query ids. Loop over it and send a message for each one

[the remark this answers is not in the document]
sharadregoti: or can i make a variable which make sure that the go func gets executed only once
isOnlyOnce = true 
if isOnlyOnce {
isOnlyOnce = false
go func 
}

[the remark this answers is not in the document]
YourTechBud: You'll need to figure out a way for this.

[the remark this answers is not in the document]
sharadregoti: range over ticker.C is not possible, only select can be done.
if we use range then go routine cannot be closed even if the ticker is stopped

utils/server/websocket.go
```diff
@@ -247,58 +267,75 @@ func (s *Server) handleGraphqlSocket(adminMan *admin.Manager) http.HandlerFunc {
...
 				data := new(model.RealtimeRequest)
 				data.ID = m.ID
 				data.Group = group.(string)
 
-				s.realtime.Unsubscribe(m.ID, data)
-				channel <- (&graphqlMessage{ID: m.ID, Payload: payloadObject{Status: utils.GqlStop}})
-
+				s.realtime.Unsubscribe(clientID, data)
+				channel <- (&graphqlMessage{ID: m.ID, Payload: payloadObject{Status: utils.GQL_STOP}})
+				graphqlIDMapper.Delete(m.ID)
```
sharadregoti: should I delete the id after unsubscribe

utils/server/websocket.go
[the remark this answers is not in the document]
YourTechBud: Stopping the ticker automatically closes the channel right?

utils/server/websocket.go
YourTechBud: Yes

utils/server/websocket.go
```diff
@@ -152,3 +164,189 @@ func (s *Server) handleWebsocket() http.HandlerFunc {
...
+
+		closeConnAliveRoutine := make(chan bool)
+		defer close(closeConnAliveRoutine)
+
+		go func() {
+			for res := range channel {
+				err := socket.WriteJSON(res)
+				if err != nil {
+					log.Println(err)
+					return
```
YourTechBud: Don't return here

utils/server/websocket.go
```diff
@@ -152,3 +164,189 @@ func (s *Server) handleWebsocket() http.HandlerFunc {
...
+				}
+			}
+		}()
+
+		for {
+
+			m := graphqlMessage{}
+			if err := socket.ReadJSON(m); err != nil {
```
YourTechBud:
```suggestion
			if err := socket.ReadJSON(&m); err != nil {
```

utils/server/websocket.go
```diff
@@ -152,3 +164,189 @@ func (s *Server) handleWebsocket() http.HandlerFunc {
...
+								return
+							case <-ticker.C:
+								channel <- &graphqlMessage{ID: m.ID, Payload: payloadObject{Status: utils.GQL_CONNECTION_KEEP_ALIVE}}
+							}
+						}
+					}()
+				}
+
+			case utils.GQL_START:
+				data := new(model.RealtimeRequest)
```
YourTechBud: Can you create data below and assign all its fields while creating it?

utils/server/websocket.go
```diff
@@ -152,3 +164,189 @@ func (s *Server) handleWebsocket() http.HandlerFunc {
...
+					channel <- &graphqlMessage{ID: feed.QueryID, Payload: payloadObject{Status: utils.GQL_DATA, Data: graphqlSuccessData{Type: feed.Type, Doc: feed.Payload, DocID: feed.DocID}}}
+				})
+
+				if err != nil {
+					channel <- &graphqlMessage{ID: m.ID, Payload: payloadObject{Status: utils.GQL_ERROR, Error: []gqlError{{Message: err.Error()}}}}
+					closeConnAliveRoutine <- true
+					return
+				}
+
+				channel <- (&graphqlMessage{ID: m.ID, Payload: payloadObject{Status: utils.GQL_START, Data: feedData}})
```
YourTechBud: you can't just pass feedData like this. You'll have to iterate over it and make a `graphqlSuccessData` for each one of them and pass that in the channel in the loop itself

PR #423, opened 2019-10-03
added delete endpoint

sharadregoti:
added delete endpoint to space cloud

tested with mongo, SQL, Postgres

returns empty object if the operation is successful otherwise an error

### DIFF
--- a/modules/crud/mgo/delete.go
+++ b/modules/crud/mgo/delete.go
@@ -3,6 +3,7 @@ package mgo
 import (
 	"context"
 	"errors"
+	"go.mongodb.org/mongo-driver/mongo/options"
 
 	"github.com/spaceuptech/space-cloud/model"
 	"github.com/spaceuptech/space-cloud/utils"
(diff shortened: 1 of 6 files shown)


modules/crud/mgo/delete.go
```diff
@@ -31,3 +31,8 @@ func (m *Mongo) Delete(ctx context.Context, project, col string, req *model.Dele
 
 	return nil
 }
+
+// DeleteCollection removes a collection from database`
+func (m Mongo) DeleteCollection(ctx context.Context, project, col string) error {
+	return 	m.client.Database(project).Drop(ctx)
```
YourTechBud: You are dropping the entire database here. We need to drop just a database

modules/crud/sql/delete.go
```diff
@@ -45,3 +45,9 @@ func (s *SQL) generateDeleteQuery(ctx context.Context, project, col string, req
 	sqlString = strings.Replace(sqlString, "\"", "", -1)
 	return sqlString, args, nil
 }
+
+func (s *SQL) DeleteCollection(ctx context.Context, project, col string) error {
+	query := "DROP TABLE " + project + "." + col
+	_, err := s.client.ExecContext(ctx, query, []interface{}{}...)
```
YourTechBud: This works right?

sharadregoti:
yes, I have tested the above endpoint with the help of postman for all the databases.

PR #440, opened 2019-10-15
added types for mssql

khannaabhi:


modules/crud/crud.go
```diff
@@ -57,7 +57,8 @@ func initBlock(dbType utils.DBType, enabled bool, connection string) (Crud, erro
 
 	case utils.MySQL, utils.Postgres:
 		return sql.Init(dbType, enabled, connection)
-
```
sharadregoti: no need to create another case
just add utils.MsSQL in the above case after utils.Postgres

modules/crud/sql/sql.go
```diff
@@ -32,6 +33,9 @@ func Init(dbType utils.DBType, enabled bool, connection string) (s *SQL, err err
 
 	case utils.MySQL:
 		s.dbType = "mysql"
+	
+	case utils.SqlServer:
+		s.dbType = "mssql"
```
sharadregoti: change mssql to sqlserver

PR #466, opened 2019-10-24
added a middleware in server

sharadregoti:
used datacounter to count the incoming bytes in request & outgoing bytes in response

### DIFF
--- a/utils/handlers/metric_middleware.go
+++ b/utils/handlers/metric_middleware.go
@@ -2,11 +2,8 @@ package handlers
 
 import (
 	"github.com/gorilla/mux"
-	"github.com/miolini/datacounter"
 	"github.com/spaceuptech/space-cloud/utils/metrics"
-	"io"
 	"net/http"
-	"sync/atomic"
 )
 
 func HandleMetricMiddleWare(next http.Handler, metrics *metrics.Module) http.Handler {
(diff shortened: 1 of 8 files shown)


utils/handlers/metric_middleware.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func HandleMetricMiddleWare(next http.Handler, metrics *metrics.Module) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		projectID, ok := vars["project"]
+		if ok {
+			readerCounter := datacounter.NewReaderCounter(r.Body)
+			writerCounter := datacounter.NewResponseWriterCounter(w)
+			next.ServeHTTP(writerCounter, r)
```
YourTechBud: You'll need to overwrite `r.Body` before passing it

utils/metrics/operations.go
```diff
@@ -16,7 +16,10 @@ func (m *Module) AddEgress(project string, bytes uint64) {
 	if !m.config.IsEnabled {
 		return
 	}
-
+	if !m.config.DisableBandwidth {
```
YourTechBud:
```suggestion
	if m.config.DisableBandwidth {
```

utils/metrics/operations.go
```diff
@@ -32,6 +35,9 @@ func (m *Module) AddIngress(project string, bytes uint64) {
 	if !m.config.IsEnabled {
 		return
 	}
+	if !m.config.DisableBandwidth {
```
YourTechBud:
```suggestion
	if m.config.DisableBandwidth {
```

PR #470, opened 2019-10-26
Id's will be inserted on empty

### LINKED ISSUE #453: Schema suggestions
After going through the entire schema thing, I had few thoughts for better semantics and flexibility. I have listed them below:

- The `ID` type should automatically generate a random, unique UUID string if not provided by the mutation.
-  A table/collection should have the flexibility to use multiple fields with type `ID`. Type `ID` should not be looked as `primary key`, since primary key is a constraint and not a type. 
- The `@id` directive should be replaced by `@primary` for better semantics. This directive should be used to put Primary key constraint on a particular field.

sharadregoti:
fixed #453 
need to be tested


modules/auth/schema/schema.go
```diff
@@ -285,7 +286,11 @@ func (s *Schema) checkType(value interface{}, fieldValue *schemaFieldType) (inte
 				return nil, errors.New("String Wrong Date-Time Format")
 			}
 			return unitTimeInRFC3339, nil
-		case typeID, typeString, typeJoin:
+		case typeID:
+			if value == "" {
+				return ksuid.New().String(), nil
+			}
```
YourTechBud: Shouldn't you return value here?

modules/auth/schema/helpers.go
```diff
@@ -39,7 +39,7 @@ func checkErrors(realFieldStruct *schemaFieldType) error {
 	if realFieldStruct.Directive == directiveRelation && realFieldStruct.Kind != typeJoin {
 		return errors.New("schema : directive relation should contain user defined type got " + realFieldStruct.Kind)
 	}
-	if realFieldStruct.Kind == typeID && realFieldStruct.Directive != directiveId {
+	if realFieldStruct.Kind == typeID && realFieldStruct.Directive != directivePrimary {
 		return errors.New("schema : directive id should have type id")
```
YourTechBud:
```suggestion
		return errors.New("schema : directive primary should have type id")
```

utils/syncman/operations.go
```diff
@@ -92,9 +92,8 @@ func (s *Manager) SetProjectConfig(project *config.Project) error {
 
 	s.setProjectConfig(project)
 	if err := s.cb(s.projectConfig); err != nil {
-		return err
+		//return err
```
YourTechBud: Revert this

PR #504, opened 2019-11-12
Made sure body is closed in each handler before the handler func returns

### LINKED ISSUE #502: Use a time bound context for each handler
Currently some modules set a time bound context internally. The other modules (the ones related to config) do not use a context at all. 

We must create a time bound context in the handler function itself and pass it down to each module to make sure each handler function exits.

YourTechBud:
We still need to get all the functions in syncman to use `context`. Partially fixes #502

### DIFF
--- a/utils/handlers/config_file.go
+++ b/utils/handlers/config_file.go
@@ -17,15 +17,16 @@ func HandleSetFileStore(adminMan *admin.Manager, syncMan *syncman.Manager) http.
 		// Get the JWT token from header
 		token := utils.GetTokenFromHeader(r)
 
+		value := new(config.FileStore)
+		json.NewDecoder(r.Body).Decode(&value)
+		defer r.Body.Close()
+
 		// Check if the request is authorised
 		if err := adminMan.IsTokenValid(token); err != nil {
 			w.WriteHeader(http.StatusUnauthorized)
 			json.NewEncoder(w).Encode(map[string]string{"error": err.Error()})
 			return
 		}
-
-		value := new(config.FileStore)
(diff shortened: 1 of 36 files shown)


utils/handlers/config_file.go
```diff
@@ -17,16 +17,17 @@ func HandleSetFileStore(adminMan *admin.Manager, syncMan *syncman.Manager) http.
 		// Get the JWT token from header
 		token := utils.GetTokenFromHeader(r)
 
+		value := new(config.FileStore)
+		json.NewDecoder(r.Body).Decode(&value)
```
sharadregoti:
```suggestion
		json.NewDecoder(r.Body).Decode(value)
```

utils/handlers/config_file.go
```diff
@@ -84,14 +86,16 @@ func HandleSetFileRule(adminMan *admin.Manager, syncMan *syncman.Manager) http.H
 		// Get the JWT token from header
 		token := utils.GetTokenFromHeader(r)
 
+		value := new(config.FileRule)
+		json.NewDecoder(r.Body).Decode(&value)
```
sharadregoti:
```suggestion
		json.NewDecoder(r.Body).Decode(value)
```

utils/handlers/config_file.go
```diff
@@ -117,14 +121,16 @@ func HandleDeleteFileRule(adminMan *admin.Manager, syncMan *syncman.Manager) htt
...
```
sharadregoti:
```suggestion
		json.NewDecoder(r.Body).Decode(value)
```

PR #510, opened 2019-11-18
Implement start of Space-Cloud-Instance with generated config

### LINKED ISSUE #65: Can space-cloud download the example and run them directly?
It would be great if the `space-cloud init` command could setup the examples and start them as well.

ledex:
This fixes #65. I don't know if this is still needed, it might be better, that the initialisation of the config takes place in the web-UI.

### DIFF
--- a/config/generate.go
+++ b/config/generate.go
@@ -42,7 +42,7 @@ func generateAdmin() *Admin {
 }
 
 // GenerateConfig started the interactive cli to generate config file
-func GenerateConfig(configFilePath string) error {
+func GenerateConfig(configFilePath string) (string, error) {
 	fmt.Println()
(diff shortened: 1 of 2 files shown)


YourTechBud:
Hey ledex. This feature should be implemented in `mission-control` in my opinion. So space-cloud can be started without a config and the user can visually select a config from a bunch of templates? We already kinda do that while the user creates a new project.

What do you think?

ledex:
Jep @YourTechBud. Totally agree 👍 
I'll close this issue then and create one for mission-control.

YourTechBud:
@ledex can you tag the original issue (#65) in the issue you create in `mission-control` will be easier to keep track.

Issue #429, opened 2019-10-05
Schema validator does not throw error when nested object is not defined in schema


YourTechBud:
For the schema given below:
```graphql
type posts {
  _id: ID! @id
  title: String
  post: PostObj
}
```

`space-cloud` does not throw an error even if PostObj is not defined in the schema.

YourTechBud:
@sharadregoti You think you can get this fixed?

sharadregoti:
yes

YourTechBud:
resolved in `v0.14.0`

PR #520, opened 2019-11-27
auto imply relations from schema module.

### LINKED ISSUE #494: Auto imply nested queries from relations in schema SDL
Right now, user has to manually join related types in GraphQL queries like these:
```graphql
query {
  author @postgres{
    id
    name
    posts(where: {id: ""author.id"}) @postgres{
      title
    }
  }
}
```

It can be great if we can imply these joins from the schema. For example:

```graphql
type author {
  id: ID! @primary
  name: String!
  posts: [post] @relation
}

type post {
  id: ID! @primary
  title: String!
  user_id: author! @foreign
}
```

The `@relation` in above schema won't create any column but is mentioned so that such queries are possible:

```graphql
query {
  author @postgres{
    id
    name
    posts {
      title
    }
  }
}
```

While, the `@foreign` will create a column as well as a foreign key on the the column. The parent table key will be `id/_id` by default. But can be changed by the `to` argument like this:

```graphql
type post {
  id: ID! @primary
  title: String!
  user_id: author! @foreign(to: "name")
}
```

Similarly `@relation` directive will have it's default `to` argument as `id/_id` which can be overwritten.  For cross database joins, user can also mention db argument like these:

```graphql
type author {
  id: ID! @primary
  name: String!
  posts: [post] @relation(db: "mongo", to: "name")
}
```

> Note: This will be a breaking change. However, it will provide more clarity and transparency to user as to whether he is creating a foreign key or just creating a relation which can be used in his queries.

YourTechBud:
Fixes #494

### DIFF
--- a/modules/crud/sql/create.go
+++ b/modules/crud/sql/create.go
@@ -21,6 +21,7 @@ func (s *SQL) Create(ctx context.Context, project, col string, req *model.Create
 	if err != nil {
 		return 0, err
 	}
+
 	res, err := doExecContext(ctx, sqlQuery, args, s.client)
 	if err != nil {
 		return 0, err
--- a/modules/schema/creation.go
+++ b/modules/schema/creation.go
@@ -10,7 +10,7 @@ import (
 
 type creationModule struct {
 	dbType, project, ColName, FieldKey, columnType string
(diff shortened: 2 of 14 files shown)


YourTechBud:
This also fixes #429

PR #549, opened 2019-12-16
intent mechanism in filestore module

sharadregoti:


modules/eventing/crud.go
```diff
@@ -230,7 +232,9 @@ func (m *Module) processCreateDocs(token int, batchID, dbType, col string, rows
 	if err != nil {
 		return nil
 	}
-	rules := m.getMatchingRules(utils.EventCreate, map[string]string{"col": col, "db": dbType})
+
+	rules := m.getMatchingRules(utils.EventDBCreate, map[string]string{})
```
YourTechBud: Don't change the options here. We need the options for database operations

modules/eventing/file.go
```diff
@@ -0,0 +1,125 @@
...
+
+	rules := m.getMatchingRules(utils.EventFileCreate, map[string]string{})
+
+	// Process the documents
+	eventDocs := make([]*model.EventDocument, 0)
+	for _, rule := range rules {
+		eventDocs = append(eventDocs, m.generateQueueEventRequest(token, rule.Retries,
+			batchID, utils.EventStatusIntent, rule.Url, &model.QueueEventRequest{
+				Type:    utils.EventFileCreate,
+				Payload: req.Path,
```
YourTechBud: Payload should be a struct containing path, meta, type

modules/eventing/file.go
```diff
@@ -0,0 +1,125 @@
...
+
+	rules := m.getMatchingRules(utils.EventFileDelete, map[string]string{})
+
+	// Process the documents
+	eventDocs := make([]*model.EventDocument, 0)
+	for _, rule := range rules {
+		eventDocs = append(eventDocs, m.generateQueueEventRequest(token, rule.Retries,
+			batchID, utils.EventStatusIntent, rule.Url, &model.QueueEventRequest{
+				Type:    utils.EventFileDelete,
+				Payload: path,
```
YourTechBud: Put the right payload

modules/eventing/file.go
```diff
@@ -0,0 +1,125 @@
...
+	createRequest := &model.CreateRequest{Document: convertToArray(eventDocs), Operation: utils.All}
+	if err := m.crud.InternalCreate(ctx, m.config.DBType, m.project, m.config.Col, createRequest); err != nil {
+		return nil, errors.New("eventing module couldn't log the request - " + err.Error())
+	}
+
+	return &model.EventIntent{BatchID: batchID, Token: token, Docs: eventDocs}, nil
+}
+
+// HookDBStage stages the event so that it can be processed
+func (m *Module) HookFileStage(ctx context.Context, intent *model.EventIntent, err error) {
```
YourTechBud: Can't we use `HandleStage` instead of this? Do we really need to create a new function?

modules/eventing/handle_intents.go
```diff
@@ -157,7 +158,43 @@ func (m *Module) processIntent(eventDoc *model.EventDocument) {
...
 		if err := m.crud.InternalUpdate(ctx, m.config.DBType, m.project, m.config.Col, m.generateStageEventRequest(eventID)); err == nil {
 			// Broadcast the event so the concerned worker can process it immediately
-			eventDoc.Status = utils.EventStatusProcessed
+			eventDoc.Status = utils.EventStatusStaged
+			m.transmitEvents(eventDoc.Token, []*model.EventDocument{eventDoc})
+		}
+	case utils.EventFileCreate:
+
+		// Check if document exists in database
+		if err := m.fileStore.DoesExists(eventDoc.Payload.(string)); err != nil {
```
YourTechBud: Payload is struct. It will contain a field URL

modules/filestore/amazons3/raw.go
```diff
@@ -0,0 +1,23 @@
...
+	input := &s3.GetObjectInput{
+		Bucket: aws.String(a.bucket),
+		Key:    aws.String(path),
+	}
+
+	result, err := svc.GetObject(input)
+	if err != nil {
+		return err
+	}
+	fmt.Println(result)
```
YourTechBud: Don't print here

modules/filestore/amazons3/raw.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func (a *AmazonS3) DoesExists(path string) error {
+	svc := s3.New(a.client)
+	input := &s3.GetObjectInput{
+		Bucket: aws.String(a.bucket),
+		Key:    aws.String(path),
+	}
+
+	result, err := svc.GetObject(input)
```
YourTechBud: This will get the entire object right?

modules/filestore/gcpstorage/raw.go
```diff
@@ -0,0 +1,12 @@
+package gcpstorage
+
+import (
+	"context"
+)
+
+func (g *GCPStorage) DoesExists(path string)  error{
+	if _, err := g.client.Bucket(g.bucket).Object(path).Attrs(context.TODO()); err != nil {
```
YourTechBud: This get the entire object right?

modules/filestore/local/raw.go
```diff
@@ -0,0 +1,13 @@
+package local
+
+import "os"
+
+func (l *Local) DoesExists(path string) ( error) {
+	// check if file / folder exists
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		// path does not exist
+		return  nil
```
YourTechBud: Return err

modules/filestore/local/raw.go
```diff
@@ -0,0 +1,13 @@
+package local
+
+import "os"
+
+func (l *Local) DoesExists(path string) ( error) {
+	// check if file / folder exists
+	if _, err := os.Stat(path); os.IsNotExist(err) {
```
YourTechBud:
```suggestion
	if _, err := os.Stat(path);  err != nil {
```

modules/filestore/operations.go
```diff
@@ -26,12 +26,19 @@ func (m *Module) CreateDir(ctx context.Context, project, token string, req *mode
...
+	intent, err := m.event.HookFileCreateIntent(ctx, req)
 	if err != nil {
 		return 500, err
-	} else {
-		return 200, nil
 	}
+
+	err = m.store.CreateDir(req)
+	if err == nil {
+		return 200, err
```
YourTechBud: Need to call the stage hook before this as well

modules/filestore/operations.go
```diff
@@ -50,12 +57,19 @@ func (m *Module) DeleteFile(ctx context.Context, project, token, path string) (i
...
-	err = m.store.DeleteFile(path)
+	intent, err := m.event.HookFileDeleteIntent(ctx, path)
 	if err != nil {
-		return http.StatusInternalServerError, err
+		return 0, err
+	}
+
+	err = m.store.DeleteFile(path)
+	if err == nil {
+		return http.StatusOK, err
```
YourTechBud: Need to call the stage hook before this as we

modules/filestore/operations.go
```diff
@@ -98,11 +112,19 @@ func (m *Module) UploadFile(ctx context.Context, project, token string, req *mod
...
+	intent, err := m.event.HookFileCreateIntent(ctx, req)
 	if err != nil {
-		return http.StatusInternalServerError, err
+		return 500, err
 	}
-	return http.StatusOK, nil
+
+	err = m.store.CreateFile(req, reader)
+	if err == nil {
+		return http.StatusOK, err
```
YourTechBud: Ca thr hook before this as well

modules/filestore/store.go
```diff
@@ -20,6 +22,8 @@ type Module struct {
 	store   FileStore
 	enabled bool
 	auth    *auth.Module
+	crud    *crud.Module
```
YourTechBud: Do we really need this here?

modules/eventing/eventing.go
```diff
@@ -29,6 +30,7 @@ type Module struct {
 	functions *functions.Module
 	adminMan  *admin.Manager
 	syncMan   *syncman.Manager
+	fileStore *filestore.Module
```
YourTechBud: Do we really need this here?

modules/filestore/store.go
```diff
@@ -93,3 +99,7 @@ func initBlock(fileStoreType utils.FileStoreType, connection, endpoint, bucket s
 		return nil, utils.ErrInvalidParams
 	}
 }
+
+func (m *Module) DoesExists(path string) error {
```
YourTechBud: Put this in operations.go
- check if file storage is enabled
- check if the operation is authorised (check for read op)
- acquire a RLock before the actual operation

utils/server/server.go
```diff
@@ -75,11 +75,11 @@ func New(nodeID, clusterID string, isConsulEnabled, removeProjectScope bool, met
 	e := eventing.New(a, c, fn, adminMan, syncMan)
 
 	c.SetHooks(&model.CrudHooks{
```
YourTechBud: Pass interface which the eventing module implements. This interface will only have the `Hook` functions

modules/filestore/store.go
```diff
@@ -20,6 +22,8 @@ type Module struct {
 	store   FileStore
 	enabled bool
 	auth    *auth.Module
+	crud    *crud.Module
+	event   *eventing.Module
```
YourTechBud: Where are you setting this variable? Should it be received in the `Init` function?
Also we'll use the interface instead of the actual eventing module

modules/eventing/file.go
sharadregoti: yes we can

modules/filestore/amazons3/raw.go
sharadregoti: amazon s3 still need to be tested, i was not able to create an account.

modules/filestore/store.go
sharadregoti: no

modules/eventing/eventing.go
sharadregoti: yes, how can i call doesExist function defined in filestore module?

YourTechBud: Got it

modules/filestore/store.go
sharadregoti: even though i am not able to import use event field in filestore module, as it give rise to import cycle not allowed error

Screenshot from 2019-12-17 22-43-02

sharadregoti: so how should i call eventing functions in filestore module

YourTechBud: Make an interface!!!!

PR #550, opened 2019-12-16
remove count++ logic from schema module

sharadregoti:


YourTechBud:
We'll merge this once @khannaabhi is done with writing tests for the schema module

PR #556, opened 2019-12-17
Added default constraint to one table column.

rushikesh-kulkarni:


modules/crud/sql/helpers.go
```diff
@@ -95,6 +96,10 @@ func (s *SQL) getDBName(project, col string) string {
 	return project + "." + col
 }
 
+func (s *SQL) generateQuerySQLServer(query string) string {
+	return strings.Replace(query, "?", "@p1", -1)
```
YourTechBud: Replace $ with p1

modules/schema/schema.go
```diff
@@ -147,6 +147,17 @@ func getCollectionSchema(doc *ast.Document, dbName, collectionName string) (Sche
...
+						fieldTypeStuct.IsDefault = true
+
+						for _, arg := range directive.Arguments {
+							switch arg.Name.Value {
+							case "value":
+								val, _ := utils.ParseGraphqlValue(arg.Value, nil)
+
+								fieldTypeStuct.Default = val
+							}
+						}
```
YourTechBud: Throw error if value isn't set after the loop ends

utils/handlers/config_crud.go
```diff
@@ -217,7 +217,7 @@ func HandleModifySchema(adminMan *admin.Manager, schemaArg *schema.Schema, syncm
 		col := vars["col"]
 
 		// Create a context of execution
-		ctx, cancel := context.WithTimeout(r.Context(), 5*time.Second)
+		ctx, cancel := context.WithTimeout(r.Context(), 50*time.Second)
```
YourTechBud: Make it 10 seconds

utils/handlers/config_crud.go
```diff
@@ -399,7 +399,7 @@ func HandleModifyAllSchema(adminMan *admin.Manager, schemaArg *schema.Schema, sy
 		dbType := vars["dbType"]
 		project := vars["project"]
 		// Create a context of execution
-		ctx, cancel := context.WithTimeout(r.Context(), 10*time.Second)
+		ctx, cancel := context.WithTimeout(r.Context(), 100*time.Second)
```
YourTechBud: Make it 10 seconds

utils/handlers/graphql.go
```diff
@@ -41,7 +41,7 @@ func HandleGraphQLRequest(graphql *graphql.Module) http.HandlerFunc {
 
 		ch := make(chan struct{}, 1)
 
-		graphql.ExecGraphQLQuery(ctx, &req, token, func(op interface{}, err error) {
+		go graphql.ExecGraphQLQuery(ctx, &req, token, func(op interface{}, err error) {
```
YourTechBud: Not a go routine

modules/schema/schema.go
sharadregoti: @YourTechBud i think we should check if the default field is set or not. which is an interface after the loop ends.
by the way where is the **value** filled is created/initialized

PR #561, opened 2019-12-19
Schema validation tests

suyash16999:


modules/schema/validatecreate_test.go
```diff
@@ -0,0 +1,383 @@
...
+
+	tdd := []struct {
+		dbName, coll, description string
+		value                     model.CreateRequest
+		want                      error
+	}{
+		{
+			dbName: "sqlserver",
+			coll:   "tweet",
+			want:   errors.New("No db was found named sqlserver"),
```
sharadregoti: change in every case want accordingly to bool,
if the data provided in test case is such that error will occur then make it false otherwise true

sharadregoti: remove the hard coding of error in all the test cases

modules/schema/validatecreate_test.go
```diff
@@ -0,0 +1,383 @@
...
+
+	temp := crud.Module{}
+	s := Init(&temp, false)
+	err := s.parseSchema(Parsedata)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, val := range tdd {
+		t.Run(val.description, func(t *testing.T) {
```
sharadregoti:
```suggestion
		t.Run(val.name, func(t *testing.T) {
```

modules/schema/validatecreate_test.go
```diff
@@ -0,0 +1,383 @@
...
+			err := s.ValidateCreateOperation(val.dbName, val.coll, &val.value)
+			if !reflect.DeepEqual(err, val.want) {
```
sharadregoti: change this condition to something like this
```suggestion
			if (err != nil) != val.isErrExpected {
```

modules/schema/validatecreate_test.go
```diff
@@ -0,0 +1,383 @@
...
+			coll:   "tweet",
+			want:   errors.New("No db was found named sqlserver"),
+			value: model.CreateRequest{
+				Document: map[string]interface{}{
+					"male": true,
+				},
+			},
+		},
+		{
+			dbName: "mongo",
```
sharadregoti: name/ description is not provided in any of the test cases.

YourTechBud:
@suyash16999, @sharadregoti resolve the merge conflicts

modules/schema/schema_test.go
```diff
@@ -0,0 +1,149 @@
...
+		},
+	}
+
+	s := Init(&crud.Module{}, false)
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			r, err := s.parser(testCase.Data)
+			if (err != nil) != testCase.IsErrExpected {
+				t.Errorf("\n Schema.parseSchema() error = expected error-%v,got error-%v", testCase.IsErrExpected, err)
+			} else if !reflect.DeepEqual(r, testCase.schema) {
```
sharadregoti: remove else & add return statement in case of error

modules/schema/validatecreate_test.go
```diff
@@ -0,0 +1,442 @@
...
+	if err != nil {
+		t.Errorf("Error while parsing schema:%v", err)
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			result, err := s.schemaValidator(testCase.coll, s.SchemaDoc["mongo"][testCase.coll], testCase.Document)
+			if (err != nil) != testCase.IsErrExpected {
+				t.Errorf("\n SchemaValidateOperation() error : expected error-%v, got-%v)", testCase.IsErrExpected, err)
+			}
+			if testCase.IsSkipable == false {
```
sharadregoti:
```suggestion
			if !testCase.IsSkipable {
```

modules/schema/validatecreate_test.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+	r := s.SchemaDoc["mongo"]["tweet"]
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			for key, value := range testCase.Document {
+				retval, err := s.checkType(testCase.coll, value, r[key])
+				if (err != nil) != testCase.IsErrExpected {
+					t.Errorf("\n CheckType() error = Expected error-%v,got-%v)", testCase.IsErrExpected, err)
+				}
+				if testCase.IsSkipable == false {
```
sharadregoti:
```suggestion
				if !testCase.IsSkipable {
```

modules/schema/schema_test.go
sharadregoti:
```suggestion
			} 
			if !reflect.DeepEqual(r, testCase.schema) {
```

modules/schema/schema_test.go
```diff
@@ -0,0 +1,149 @@
...
+			},
+		},
+	}
+
+	s := Init(&crud.Module{}, false)
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			r, err := s.parser(testCase.Data)
+			if (err != nil) != testCase.IsErrExpected {
+				t.Errorf("\n Schema.parseSchema() error = expected error-%v,got error-%v", testCase.IsErrExpected, err)
```
sharadregoti:
```suggestion
				t.Errorf("\n Schema.parseSchema() error = expected error-%v,got error-%v", testCase.IsErrExpected, err)
				return
```

PR #629, opened 2020-01-11
Boltdb

sharadregoti:
Testing is remaianing

### DIFF
--- a/gateway/go.mod
+++ b/gateway/go.mod
@@ -4,7 +4,6 @@ require (
 	cloud.google.com/go v0.38.0
 	github.com/armon/go-metrics v0.0.0-20190430140413-ec5e00d3c878 // indirect
 	github.com/aws/aws-sdk-go v1.19.7
-	github.com/boltdb/bolt v1.3.1
 	github.com/coreos/bbolt v1.3.3 // indirect
 	github.com/coreos/etcd v3.3.17+incompatible
 	github.com/coreos/go-semver v0.3.0 // indirect
@@ -54,7 +53,7 @@ require (
 	github.com/xdg/scram v0.0.0-20180814205039-7eeb5667e42c // indirect
 	github.com/xdg/stringprep v1.0.0 // indirect
 	github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2 // indirect
-	go.etcd.io/bbolt v1.3.3 // indirect
+	go.etcd.io/bbolt v1.3.3
 	go.mongodb.org/mongo-driver v1.1.1
 	go.opencensus.io v0.22.0 // indirect
 	go.uber.org/zap v1.13.0 // indirect
--- a/gateway/go.mod
+++ b/gateway/go.mod
@@ -4,6 +4,7 @@ require (
 	cloud.google.com/go v0.38.0
 	github.com/armon/go-metrics v0.0.0-20190430140413-ec5e00d3c878 // indirect
 	github.com/aws/aws-sdk-go v1.19.7
+	github.com/boltdb/bolt v1.3.1
 	github.com/coreos/bbolt v1.3.3 // indirect
 	github.com/coreos/etcd v3.3.17+incompatible
 	github.com/coreos/go-semver v0.3.0 // indirect
--- a/gateway/modules/crud/bolt/aggregate.go
+++ b/gateway/modules/crud/bolt/aggregate.go
@@ -2,10 +2,11 @@ package bolt
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/spaceuptech/space-cloud/gateway/model"
 )
 
 func (b *Bolt) Aggregate(ctx context.Context, project, col string, req *model.AggregateRequest) (interface{}, error) {
-	return nil, nil
+	return nil, fmt.Errorf("aggregate operation not supported for selected database")
 }
--- a/gateway/modules/crud/bolt/aggregate.go
+++ b/gateway/modules/crud/bolt/aggregate.go
@@ -1,0 +1,11 @@
+package bolt
+
+import (
+	"context"
+
+	"github.com/spaceuptech/space-cloud/gateway/model"
+)
+
+func (b *Bolt) Aggregate(ctx context.Context, project, col string, req *model.AggregateRequest) (interface{}, error) {
+	return nil, nil
+}
--- a/gateway/modules/crud/bolt/batch.go
+++ b/gateway/modules/crud/bolt/batch.go
@@ -1,0 +1,11 @@
+package bolt
+
+import (
+	"context"
+
+	"github.com/spaceuptech/space-cloud/gateway/model"
+)
+
+func (b *Bolt) Batch(ctx context.Context, project string, req *model.BatchRequest) ([]int64, error) {
+	return nil, nil
+}
--- a/gateway/modules/crud/bolt/batch.go
+++ b/gateway/modules/crud/bolt/batch.go
@@ -2,10 +2,11 @@ package bolt
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/spaceuptech/space-cloud/gateway/model"
 )
 
 func (b *Bolt) Batch(ctx context.Context, project string, req *model.BatchRequest) ([]int64, error) {
-	return nil, nil
+	return nil, fmt.Errorf("batch operation not supported for selected database")
 }
--- a/gateway/modules/crud/bolt/bolt.go
+++ b/gateway/modules/crud/bolt/bolt.go
@@ -68,5 +68,5 @@ func (b *Bolt) connect() error {
 
 // GetDBAlias returns the dbType of the crud block
 func (b *Bolt) GetDBType() utils.DBType {
-	return utils.BoltDb
+	return utils.BoltDB
 }
--- a/gateway/modules/crud/bolt/bolt.go
+++ b/gateway/modules/crud/bolt/bolt.go
@@ -1,8 +1,11 @@
 package bolt
 
 import (
-	"github.com/boltdb/bolt"
+	"fmt"
+	"log"
+
 	"github.com/sirupsen/logrus"
+	"go.etcd.io/bbolt"
 
 	"github.com/spaceuptech/space-cloud/gateway/utils"
 )
@@ -10,15 +13,15 @@ import (
 type Bolt struct {
 	enabled    bool
 	connection string
-	client     *bolt.DB
+	client     *bbolt.DB
 }
 
 // Init initialises a new bolt instance
-func Init(enabled bool, connection string) (mongoStub *Bolt, err error) {
-	mongoStub = &Bolt{enabled: enabled, connection: connection, client: nil}
+func Init(enabled bool, connection string) (b *Bolt, err error) {
+	b = &Bolt{enabled: enabled, connection: connection}
 
-	if mongoStub.enabled {
-		err = mongoStub.connect()
+	if b.enabled {
+		err = b.connect()
 	}
 
 	return
@@ -50,11 +53,15 @@ func (b *Bolt) IsClientSafe() error {
 }
 
 func (b *Bolt) connect() error {
-	client, err := bolt.Open(b.connection, 0600, bolt.DefaultOptions)
+	if err := b.Close(); err != nil {
+		return fmt.Errorf("error closing previous database connection in bbolt db")
+	}
+
+	client, err := bbolt.Open(b.connection, 0600, bbolt.DefaultOptions)
 	if err != nil {
 		return err
 	}
-
+	log.Println("successfully connected to bbolt database")
 	b.client = client
 	return nil
 }
--- a/gateway/modules/crud/bolt/bolt.go
+++ b/gateway/modules/crud/bolt/bolt.go
@@ -1,0 +1,65 @@
+package bolt
+
+import (
+	"github.com/boltdb/bolt"
+	"github.com/sirupsen/logrus"
+
+	"github.com/spaceuptech/space-cloud/gateway/utils"
+)
+
+type Bolt struct {
+	enabled    bool
+	connection string
+	client     *bolt.DB
+}
+
+// Init initialises a new bolt instance
+func Init(enabled bool, connection string) (mongoStub *Bolt, err error) {
+	mongoStub = &Bolt{enabled: enabled, connection: connection, client: nil}
+
+	if mongoStub.enabled {
+		err = mongoStub.connect()
+	}
+
+	return
+}
+
+// Close gracefully the Bolt client
+func (b *Bolt) Close() error {
+	if b.client != nil {
+		return b.client.Close()
+	}
+
+	return nil
+}
+
+// IsClientSafe checks whether database is enabled and connected
+func (b *Bolt) IsClientSafe() error {
+	if !b.enabled {
+		return utils.ErrDatabaseDisabled
+	}
+
+	if b.client == nil {
+		if err := b.connect(); err != nil {
+			logrus.Errorf("Error connecting to bboltdb - %v", err)
+			return utils.ErrDatabaseConnection
+		}
+	}
+
+	return nil
+}
+
+func (b *Bolt) connect() error {
+	client, err := bolt.Open(b.connection, 0600, bolt.DefaultOptions)
+	if err != nil {
+		return err
+	}
+
+	b.client = client
+	return nil
+}
+
+// GetDBAlias returns the dbType of the crud block
+func (b *Bolt) GetDBType() utils.DBType {
+	return utils.BoltDb
+}
--- a/gateway/modules/crud/bolt/collections.go
+++ b/gateway/modules/crud/bolt/collections.go
@@ -2,14 +2,15 @@ package bolt
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/spaceuptech/space-cloud/gateway/utils"
 )
 
 func (b *Bolt) GetCollections(ctx context.Context, project string) ([]utils.DatabaseCollections, error) {
-	return nil, nil
+	return nil, fmt.Errorf("error getting collection operation not supported for selected database")
 }
 
 func (b *Bolt) DeleteCollection(ctx context.Context, project, col string) error {
-	return nil
+	return fmt.Errorf("error deleting collection operation not supported for selected database")
 }
--- a/gateway/modules/crud/bolt/collections.go
+++ b/gateway/modules/crud/bolt/collections.go
@@ -1,0 +1,15 @@
+package bolt
+
+import (
+	"context"
+
+	"github.com/spaceuptech/space-cloud/gateway/utils"
+)
+
+func (b *Bolt) GetCollections(ctx context.Context, project string) ([]utils.DatabaseCollections, error) {
+	return nil, nil
+}
+
+func (b *Bolt) DeleteCollection(ctx context.Context, project, col string) error {
+	return nil
+}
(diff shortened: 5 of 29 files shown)


YourTechBud:
Test the library first. Let's get this off the ground. Also, your tests seem to be failing

gateway/go.mod
```diff
@@ -4,6 +4,7 @@ require (
 	cloud.google.com/go v0.38.0
 	github.com/armon/go-metrics v0.0.0-20190430140413-ec5e00d3c878 // indirect
 	github.com/aws/aws-sdk-go v1.19.7
+	github.com/boltdb/bolt v1.3.1
```
YourTechBud: You are using the wrong library. Use `bbolt` https://github.com/etcd-io/bbolt

gateway/modules/crud/bolt/aggregate.go
```diff
@@ -0,0 +1,11 @@
+package bolt
+
+import (
+	"context"
+
+	"github.com/spaceuptech/space-cloud/gateway/model"
+)
+
+func (b *Bolt) Aggregate(ctx context.Context, project, col string, req *model.AggregateRequest) (interface{}, error) {
+	return nil, nil
```
YourTechBud: Send an error that it isn't supported

gateway/modules/crud/bolt/batch.go
```diff
@@ -0,0 +1,11 @@
+package bolt
+
+import (
+	"context"
+
+	"github.com/spaceuptech/space-cloud/gateway/model"
+)
+
+func (b *Bolt) Batch(ctx context.Context, project string, req *model.BatchRequest) ([]int64, error) {
+	return nil, nil
```
YourTechBud: Send error

gateway/modules/crud/bolt/bolt.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type Bolt struct {
+	enabled    bool
+	connection string
+	client     *bolt.DB
+}
+
+// Init initialises a new bolt instance
+func Init(enabled bool, connection string) (mongoStub *Bolt, err error) {
```
YourTechBud: why is the variable here called mongo stub?

gateway/modules/crud/bolt/bolt.go
```diff
@@ -0,0 +1,65 @@
...
+
+	return nil
+}
+
+func (b *Bolt) connect() error {
+	client, err := bolt.Open(b.connection, 0600, bolt.DefaultOptions)
+	if err != nil {
+		return err
+	}
+
```
YourTechBud: log success message here. Do this for all dbs as well

gateway/modules/crud/bolt/collections.go
```diff
@@ -0,0 +1,15 @@
+package bolt
+
+import (
+	"context"
+
+	"github.com/spaceuptech/space-cloud/gateway/utils"
+)
+
+func (b *Bolt) GetCollections(ctx context.Context, project string) ([]utils.DatabaseCollections, error) {
+	return nil, nil
```
YourTechBud: return error

gateway/modules/crud/bolt/collections.go
```diff
@@ -0,0 +1,15 @@
...
+}
+
+func (b *Bolt) DeleteCollection(ctx context.Context, project, col string) error {
+	return nil
```
YourTechBud: return error

gateway/modules/crud/bolt/create.go
```diff
@@ -0,0 +1,100 @@
...
+func (b *Bolt) Create(ctx context.Context, project, col string, req *model.CreateRequest) (int64, error) {
+	// Create a collection object
+	// todo check if exists
+	switch req.Operation {
+	case utils.One:
+		// Insert single document
+		if err := b.client.Update(func(tx *bolt.Tx) error {
+			count, _, err := b.Read(ctx, project, col, &model.ReadRequest{
+				Find:      req.Document.(map[string]interface{}),
+				Operation: utils.All,
```
YourTechBud: Use the operation count here. Will save you from all the unmarshalling overload

gateway/modules/crud/bolt/create.go
```diff
@@ -0,0 +1,100 @@
...
+// Create inserts a document (or multiple when op is "all") into the database
+func (b *Bolt) Create(ctx context.Context, project, col string, req *model.CreateRequest) (int64, error) {
+	// Create a collection object
+	// todo check if exists
+	switch req.Operation {
+	case utils.One:
+		// Insert single document
+		if err := b.client.Update(func(tx *bolt.Tx) error {
+			count, _, err := b.Read(ctx, project, col, &model.ReadRequest{
+				Find:      req.Document.(map[string]interface{}),
```
YourTechBud: Just check if the `_id` exists. Don't check for the entire document, that is incorrect. What if the user insert two documents with the same `_id` but different `name` fields

gateway/modules/crud/bolt/create.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Create inserts a document (or multiple when op is "all") into the database
+func (b *Bolt) Create(ctx context.Context, project, col string, req *model.CreateRequest) (int64, error) {
+	// Create a collection object
+	// todo check if exists
+	switch req.Operation {
+	case utils.One:
+		// Insert single document
+		if err := b.client.Update(func(tx *bolt.Tx) error {
+			count, _, err := b.Read(ctx, project, col, &model.ReadRequest{
```
YourTechBud: If you are using the `count` operation. The `count` variable will always be one, even if no documents were present. You having to probe the result in this case.

gateway/modules/crud/bolt/create.go
```diff
@@ -0,0 +1,100 @@
...
+				logrus.Error("error inserting into bboltdb data already exists - %v", err)
+			}
+
+			b, err := tx.CreateBucketIfNotExists([]byte(project))
+			if err != nil {
+				logrus.Errorf("error creating bucket in bboltdb while inserting- %v", err)
+				return err
+			}
+			id, ok := req.Document.(map[string]interface{})["_id"]
+			if ok {
```
YourTechBud: Check for the `!ok` condition and throw error there only

gateway/modules/crud/bolt/create.go
```diff
@@ -0,0 +1,100 @@
...
+				value, err := json.Marshal(req.Document)
+				if err != nil {
+					logrus.Errorf("error marshalling while inserting in bboltdb - %v", err)
+					return err
+				}
+
+				if err = b.Put([]byte(fmt.Sprintf("%s/%s", col, id)), value); err != nil {
+					return err
```
YourTechBud: Why no log error here

gateway/modules/crud/bolt/create.go
```diff
@@ -0,0 +1,100 @@
...
+				}
+			}
+			logrus.Errorf("error bboltdb unable to find _id while inserting - %v", err)
```
YourTechBud: According to your logic this will always be executed. Not good. Use the `!ok` system i suggested above

gateway/modules/crud/bolt/create.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/boltdb/bolt"
+	"github.com/sirupsen/logrus"
+
+	"github.com/spaceuptech/space-cloud/gateway/model"
+	"github.com/spaceuptech/space-cloud/gateway/utils"
+)
+
+// Create inserts a document (or multiple when op is "all") into the database
+func (b *Bolt) Create(ctx context.Context, project, col string, req *model.CreateRequest) (int64, error) {
```
YourTechBud: You have seperated out the implementation of `all` and `one` operations. I would suggest you only implement the `all` operation. Convert the object to an array if it was a `one` operation. Let's keep things DRY.

gateway/modules/crud/bolt/create.go
```diff
@@ -0,0 +1,100 @@
...
+				}
+
+				if err = b.Put([]byte(fmt.Sprintf("%s/%s", col, id)), value); err != nil {
+					return err
+				}
+			}
+			logrus.Errorf("error bboltdb unable to find _id while inserting - %v", err)
+			return err
+		}); err != nil {
+			return 0, nil
```
YourTechBud:
```suggestion
			return 0, err
```

gateway/modules/crud/bolt/create.go
```diff
@@ -0,0 +1,100 @@
...
+					return err
+				}
+
+				if err = bucket.Put([]byte(fmt.Sprintf("%s/%s", col, id)), value); err != nil {
+					return err
+				}
+			}
+			return nil
+		}); err != nil {
+			return 0, nil
```
YourTechBud:
```suggestion
			return 0, err
```

gateway/modules/crud/bolt/raw.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+// GetConnectionState : function to check connection state
+func (b *Bolt) GetConnectionState(ctx context.Context) bool {
+	if !b.enabled || b.client == nil {
+		return false
+	}
+
+	// Ping to check if connection is established
+	err := b.connect()
```
YourTechBud: Is there a better way to do this?

gateway/modules/crud/bolt/describe.go
```diff
@@ -0,0 +1,11 @@
+package bolt
+
+import (
+	"context"
+
+	"github.com/spaceuptech/space-cloud/gateway/utils"
+)
+
+func (b *Bolt) DescribeTable(ctc context.Context, project, col string) ([]utils.FieldType, []utils.ForeignKeysType, []utils.IndexType, error) {
+	return nil, nil, nil, nil
```
YourTechBud: return error

gateway/modules/crud/bolt/delete.go
```diff
@@ -0,0 +1,83 @@
...
+				}
+				// if valid then update
+				if utils.Validate(req.Find, result) {
+					// delete data
+					if err := bucket.Delete(k); err != nil {
+						return err
+					}
+					// exit the loop
+					break
+				}
```
YourTechBud: Just have one function for both operations. Break if op was one else continue

gateway/modules/crud/bolt/delete.go
```diff
@@ -0,0 +1,83 @@
...
+
+			// get all keys matching the prefix
+			prefix := []byte(col + "/")
+			for k, v := c.Seek(prefix); k != nil && bytes.HasPrefix(k, prefix); k, v = c.Next() {
+				result := map[string]interface{}{}
+				if err := json.Unmarshal(v, result); err != nil {
+					logrus.Errorf("error un marshalling while reading from bboltdb - %v", err)
+					return err
+				}
+				// if valid then update
```
YourTechBud:
```suggestion
				// if valid then delete
```

gateway/modules/crud/bolt/raw.go
```diff
@@ -0,0 +1,26 @@
+package bolt
+
+import "context"
+
+func (b *Bolt) RawExec(ctx context.Context, project string) error {
+	return nil
```
YourTechBud: return error in all unsupported operations

gateway/modules/crud/bolt/read.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/boltdb/bolt"
+	"github.com/sirupsen/logrus"
+
+	"github.com/spaceuptech/space-cloud/gateway/model"
+	"github.com/spaceuptech/space-cloud/gateway/utils"
+)
+
+func (b *Bolt) Read(ctx context.Context, project, col string, req *model.ReadRequest) (int64, interface{}, error) {
+
+	switch req.Operation {
```
YourTechBud: We can have a single implementation for one an all

gateway/modules/crud/bolt/read.go
```diff
@@ -0,0 +1,70 @@
...
+
+		var count int64
+		results := []interface{}{}
+		if err := b.client.View(func(tx *bolt.Tx) error {
+			// Assume bucket exists and has keys
+			c := tx.Bucket([]byte(project)).Cursor()
+
+			prefix := []byte(col + "/")
+			for k, v := c.Seek(prefix); k != nil && bytes.HasPrefix(k, prefix); k, v = c.Next() {
+				count++
```
YourTechBud: Don't update the count here. Update the count where you amend the the result

gateway/modules/crud/bolt/read.go
```diff
@@ -0,0 +1,70 @@
...
+					logrus.Errorf("error un marshalling while reading from bboltdb - %v", err)
+					return err
+				}
+				if utils.Validate(req.Find, result) {
+					results = append(results, result)
+				}
+			}
+			return nil
+		}); err != nil {
+			return 0, nil, nil
```
YourTechBud:
```suggestion
			return 0, nil, err
```

gateway/modules/crud/bolt/delete.go
```diff
@@ -0,0 +1,83 @@
...
+					count++
+					// delete data
+					if err := bucket.Delete(k); err != nil {
+						return err
+					}
+				}
+			}
+			return nil
+		}); err != nil {
+			return 0, nil
```
YourTechBud:
```suggestion
			return 0, err
```

gateway/modules/crud/bolt/read.go
```diff
@@ -0,0 +1,70 @@
...
+					logrus.Errorf("error un marshalling while reading from bboltdb - %v", err)
+					return err
+				}
+				if utils.Validate(req.Find, result) {
+					finalResult = result
+				}
+			}
+			return nil
+		}); err != nil {
+			return 0, nil, nil
```
YourTechBud:
```suggestion
			return 0, nil, err
```

gateway/modules/crud/bolt/update.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/spaceuptech/space-cloud/gateway/model"
+	"github.com/spaceuptech/space-cloud/gateway/utils"
+)
+
+func (b *Bolt) Update(ctx context.Context, project, col string, req *model.UpdateRequest) (int64, error) {
+
+	switch req.Operation {
+	case utils.One, utils.Upsert:
```
YourTechBud: Upsert acts as regular update `all` operation when there one or more documents already exist which match the find clause

gateway/modules/crud/bolt/update.go
```diff
@@ -0,0 +1,95 @@
...
+			prefix := []byte(col + "/")
+			for k, v := c.Seek(prefix); k != nil && bytes.HasPrefix(k, prefix); k, v = c.Next() {
+				result := map[string]interface{}{}
+				if err := json.Unmarshal(v, result); err != nil {
+					logrus.Errorf("error un marshalling while reading from bboltdb - %v", err)
+					return err
+				}
+				// if valid then update
+				if utils.Validate(req.Find, result) {
+					value, err := json.Marshal(req.Update)
```
YourTechBud: req,Update contains a map like the following:
```json
{
  "$set": {"name": "John"}
}
```
You need to selectively update fields in the current object, and overwrite using the updated object

gateway/modules/crud/bolt/update.go
```diff
@@ -0,0 +1,95 @@
...
+					if err = bucket.Put(k, value); err != nil {
+						return err
+					}
+					// exit the loop
+					break
+				}
+			}
+			return nil
+		}); err != nil {
+			return 0, nil
```
YourTechBud:
```suggestion
			return 0, err
```

gateway/modules/crud/bolt/update.go
```diff
@@ -0,0 +1,95 @@
...
+
+					// over ride the data
+					if err = bucket.Put(k, value); err != nil {
+						return err
+					}
+				}
+			}
+			return nil
+		}); err != nil {
+			return 0, nil
```
YourTechBud:
```suggestion
			return 0, err
```

gateway/modules/crud/bolt/raw.go
sharadregoti: image
we do have some utility function which can be used, if succesfully executed we can assume it is client is still connected with the database

gateway/modules/crud/bolt/create.go
sharadregoti: i don't think so,
my implementatin 
image

YourTechBud: You've implement count incorrectly. Check the implementation for other databases

gateway/modules/crud/bolt/update.go
```diff
@@ -0,0 +1,89 @@
...
+						// exit the loop
+						break
+					}
+				}
+			}
+			return nil
+		}); err != nil {
+			return 0, err
+		}
+		// todo $set should be map or array ?
```
YourTechBud: remove this comment

gateway/modules/crud/bolt/update.go
```diff
@@ -0,0 +1,89 @@
...
+		objToSet, ok := req.Update["$set"].(map[string]interface{})
+		if !ok {
+			return 0, fmt.Errorf("error unable to update in bbolt - $set db operator not found or the operator value is not map")
+		}
```
YourTechBud: put this inside the if

gateway/modules/userman/operations.go
```diff
@@ -27,7 +27,7 @@ func (m *Module) Profile(ctx context.Context, token, dbType, project, id string)
 		return 0, nil, err
 	}
 	switch utils.DBType(actualDbType) {
-	case utils.Mongo:
+	case utils.Mongo, utils.BoltDb:
```
YourTechBud: It should be `utils.BoltDB`

gateway/modules/crud/bolt/raw.go
```diff
@@ -0,0 +1,27 @@
...
+	"context"
+	"fmt"
+)
+
+func (b *Bolt) RawExec(ctx context.Context, project string) error {
+	return fmt.Errorf("error raw exec cannot be performed over selected database")
+}
+
+func (b *Bolt) CreateProjectIfNotExist(ctx context.Context, project string) error {
+	return fmt.Errorf("errocr create project operation cannot be performed over selected database")
```
YourTechBud: Error

gateway/modules/crud/bolt/read.go
```diff
@@ -0,0 +1,73 @@
...
+					if req.Operation == utils.One {
+						break
+					}
+				}
+			}
+			return nil
+		}); err != nil {
+			return 0, nil, err
+		}
+		if req.Operation == utils.One && count == 1 {
```
YourTechBud: Return error if operation was one but count is zero

gateway/utils/handlers/filestore.go
```diff
@@ -124,6 +124,16 @@ func HandleRead(auth *auth.Module, fileStore *filestore.Module) http.HandlerFunc
 			}
 			json.NewEncoder(w).Encode(map[string]interface{}{"result": res})
 			return
+		} else if op == "exist" {
+			log.Println("ldsfilh", path)
```
YourTechBud: Why is this here

gateway/utils/handlers/filestore.go
```diff
@@ -124,6 +124,16 @@ func HandleRead(auth *auth.Module, fileStore *filestore.Module) http.HandlerFunc
...
+			if err := fileStore.DoesExists(ctx, project, token, path); err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
```
YourTechBud: Return not found status code

PR #634, opened 2020-01-13
Boltdb

sharadregoti:
upserting is not very well tested, for now like utils.One operation


YourTechBud:
We need to update how count works in each package. The result needs to be a map otherwise the result won't be accessible in GraphQL

PR #640, opened 2020-01-15
Path verification

sharadregoti:


runner/server/handle.go
```diff
@@ -96,6 +98,14 @@ func (s *Server) HandleApplyService() http.HandlerFunc {
 		json.NewDecoder(r.Body).Decode(req)
 
 		if req.Data.Meta.IsDeploy {
+			// verify path e.g -> /artifacts/acc_id/projectid/env/version/build.zip
+			arr := strings.Split(req.Data.Path, "/")
+			// 5 will ensure that there will not be any index out of range error
+			if len(arr) < 5 || arr[3] != req.Data.Meta.Service.ProjectID || arr[4] != req.Data.Meta.Service.Environment || arr[5] != req.Data.Meta.Service.Environment {
```
YourTechBud: Length should be 7

runner/server/handle.go
```diff
@@ -96,6 +98,14 @@ func (s *Server) HandleApplyService() http.HandlerFunc {
 		json.NewDecoder(r.Body).Decode(req)
 
 		if req.Data.Meta.IsDeploy {
+			// verify path e.g -> /artifacts/acc_id/projectid/env/version/build.zip
+			arr := strings.Split(req.Data.Path, "/")
+			// 5 will ensure that there will not be any index out of range error
+			if len(arr) < 7 || arr[3] != req.Data.Meta.Service.ProjectID || arr[4] != req.Data.Meta.Service.Environment || arr[5] != req.Data.Meta.Service.Environment {
```
YourTechBud: Length should be exactly seven. Also update the comment above

PR #641, opened 2020-01-16
added sc storage interface for let's encrypt

sharadregoti:
for checking the existence of key, i have used getone function instead of count as the implementation differs for boltdb & other db supported by space cloud.


YourTechBud:
For checking if keys exists, you should use getAll and check if the array returned has length greater than zero

gateway/main.go
```diff
@@ -131,6 +131,21 @@ var essentialFlags = []cli.Flag{
 		Usage:  "The database / topic to push the metrics to",
 		EnvVar: "METRICS_SCOPE",
 	},
+	cli.StringFlag{
```
YourTechBud: Don't take cli flags for this. Read them directly from environment variables

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,122 @@
...
+
+	m := types.M{
+		"_id":   key,
+		"value": base64.StdEncoding.EncodeToString(value),
+		"size":  len(value),
+	}
+
+	data, err := s.db.Upsert(s.collection).Set(m).Apply(ctx)
+	if err != nil || data.Status != http.StatusOK {
+		logrus.Errorf("error while storing data of lets encrypt - %v", err)
```
YourTechBud: When status code isn't 200, data.error will gve the error message

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,122 @@
...
+func (s *storage) Load(key string) ([]byte, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	data, err := s.db.GetOne(s.collection).Where(types.M{"_id": key}).Apply(ctx)
+	if err != nil || data.Status != http.StatusOK {
+		logrus.Errorf("error while retrieving data of lets encrypt - %v", err)
+		return nil, err
+	}
+	return json.Marshal(data.Data)
```
YourTechBud: You only need to return the value here. Not the entire object.

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (s *storage) Delete(key string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	data, err := s.db.DeleteOne(s.collection).Where(types.M{"_id": key}).Apply(ctx)
```
YourTechBud: Use delete all

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,122 @@
...
+		return err
+	}
+	return nil
+}
+
+func (s *storage) Exists(key string) bool {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	data, err := s.db.GetOne(s.collection).Where(types.M{"_id": key}).Apply(ctx)
```
YourTechBud: Use get all and then check the length of the result

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil || data.Status != http.StatusOK {
+		logrus.Errorf("error while checking existence of in lets encrypt unable to specified key - %v", err)
+		return false
+	}
+	// 4 is the number of columns in table so the map should contain 4 fields
+	if len(data.Data) != 4 {
```
YourTechBud: What is this?

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,122 @@
...
+		logrus.Errorf("error while checking existence of specified key of lets encrypt")
+		return false
+	}
+	return true
+}
+
+func (s *storage) List(prefix string, recursive bool) ([]string, error) {
+	// todo verify this
+	return []string{prefix}, nil
```
YourTechBud: You need to return a list of all keys here.

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,122 @@
...
+	modifiedTime, err := time.Parse(time.RFC3339, data.Data["modified"].(string))
+	if err != nil {
+		return certmagic.KeyInfo{}, fmt.Errorf("error while generating stats in lets encrypt unable to parse string to time - %v", err)
+	}
+
+	return certmagic.KeyInfo{
+		Key:        key,
+		Modified:   modifiedTime,
+		Size:       data.Data["size"].(int64),
+		IsTerminal: false, // todo check this
```
YourTechBud:
```suggestion
		IsTerminal: true,
```

gateway/utils/letsencrypt/sc_store.go
YourTechBud: Found a problem in `space-api-go`. Data will actually have a field named result which could contain an array or map depending on the operation

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,122 @@
...
+	}, nil
+}
+
+func (s *storage) Lock(key string) error {
+	return nil
```
YourTechBud: This needs to be implemented

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (s *storage) Unlock(key string) error {
```
YourTechBud: This needs to be implemented

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	logrus.Errorf("error while checking existence in lets encrypt length less than zero")
+	return false
+}
+
+func (s *storage) List(prefix string, recursive bool) ([]string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	response, err := s.db.Get(s.collection).Where(types.Cond("_id", "regex", prefix)).Apply(ctx)
```
YourTechBud: Use a prefix query

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,283 @@
...
+		scAddr = "store.space-cloud.svc.cluster.local:4122"
+	}
+	if scDatabase == "" {
+		scDatabase = "bolt"
+	}
+	if scCollection == "" {
+		scCollection = "certificates"
+	}
+
+	return &storage{db: apigo.New(scProject, scAddr, false).DB(scDatabase), collection: scCollection}
```
YourTechBud: Set the path here

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+	data, ok := result.([]interface{})
+	if !ok {
+		logrus.Errorf("error while checking existence in lets encrypt unable to assert result to array")
+		return false
+	}
+	if len(data) > 0 {
+		return true
+	}
+	logrus.Errorf("error while checking existence in lets encrypt length less than zero")
```
YourTechBud: No need to log this

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func (s *storage) List(prefix string, recursive bool) ([]string, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	response, err := s.db.Get(s.collection).Where(types.Cond("_id", "regex", fmt.Sprintf("^%s", prefix))).Apply(ctx)
+	if err != nil || response.Status != http.StatusOK {
+		logrus.Errorf("error while listing response of lets encrypt - %v", err)
+		return nil, err
```
YourTechBud: The error can be nil event when status was not okay

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,283 @@
...
+		"_id":   key,
+		"value": base64.StdEncoding.EncodeToString(value),
+		"size":  len(value),
+	}
+
+	data, err := s.db.Upsert(s.collection).Set(m).Apply(ctx)
+	if err != nil || data.Status != http.StatusOK {
+		logrus.Errorf("error while storing data of lets encrypt - %v %v", err, data.Error)
+	}
+	return err
```
YourTechBud: The error can be nil event when status was not okay

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func (s *storage) Load(key string) ([]byte, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	response, err := s.db.GetOne(s.collection).Where(types.M{"_id": key}).Apply(ctx)
+	if err != nil || response.Status != http.StatusOK {
+		logrus.Errorf("error while retrieving response of lets encrypt - %v", err)
+		return nil, err
```
YourTechBud: The error can be nil event when status was not okay

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,283 @@
...
+
+func (s *storage) Delete(key string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	data, err := s.db.Delete(s.collection).Where(types.M{"_id": key}).Apply(ctx)
+	if err != nil || data.Status != http.StatusOK {
+		logrus.Errorf("error while deleting data of lets encrypt - %v %v", err, data.Error)
+	}
+	return err
```
YourTechBud: The error can be nil event when status was not okay

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func (s *storage) Stat(key string) (certmagic.KeyInfo, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	data, err := s.db.GetOne(s.collection).Where(types.M{"_id": key}).Apply(ctx)
+	if err != nil || data.Status != http.StatusOK {
+		logrus.Errorf("error getting stats in lets encrypt - %v", err)
+		return certmagic.KeyInfo{}, err
```
YourTechBud: The error can be nil event when status was not okay

gateway/utils/letsencrypt/sc_store.go
```diff
@@ -0,0 +1,311 @@
...
+func NewScStore() *storage {
+	scProject := os.Getenv("LETSENCRYPT_SC_PROJECT")
+	scAddr := os.Getenv("LETSENCRYPT_SC_ADDR")
+	scDatabase := os.Getenv("LETSENCRYPT_SC_DATABASE")
+	scCollection := os.Getenv("LETSENCRYPT_SC_COLLECTION")
+	if scProject == "" {
+		scProject = "space_cloud"
+	}
+	if scAddr == "" {
+		scAddr = "store.space-cloud.svc.cluster.local:4122"
```
YourTechBud:
```suggestion
		scAddr = "store.space_cloud.svc.cluster.local:4122"
```

PR #661, opened 2020-01-25
Implementing kube secrets

high-functioning-sociopath:


runner/model/service.go
```diff
@@ -15,6 +15,7 @@ type Service struct {
 	Upstreams   []Upstream        `json:"upstreams" yaml:"upstreams"`
 	Runtime     Runtime           `json:"runtime" yaml:"runtime"`
 	Expose      *Expose           `json:"expose" yaml:"expose"`
+	Secrets     []string          `json:"secrets" yaml:"secrets"`
```
YourTechBud: Secrets should be on the task level

runner/utils/driver/istio/istio_helpers.go
```diff
@@ -20,44 +20,94 @@ import (
...
-		ports := prepareContainerPorts(task.Ports)
-
-		// Prepare command and args
-		var cmd, args []string
-		if task.Docker.Cmd != nil {
-			cmd = task.Docker.Cmd[0:1]
-			if len(task.Docker.Cmd) > 1 {
-				args = task.Docker.Cmd[1:]
+	// Prepare Volume
+	volume := make([]v1.Volume, len(service.Secrets))
```
YourTechBud: Are you sure the length will be equal to length of secrets array? What if all the secrets are envs?

runner/utils/driver/istio/istio_helpers.go
```diff
@@ -20,44 +20,94 @@ import (
...
+	for j, task := range tasks {
+		for _, secretName := range service.Secrets {
```
YourTechBud: Wait why is everything inside this loop? That's completely wrong. You'll be adding the same container multiple times cause of this

runner/utils/driver/istio/istio_helpers.go
```diff
@@ -20,44 +20,94 @@ import (
...
+			if task.Docker.Cmd != nil {
+				cmd = task.Docker.Cmd[0:1]
+				if len(task.Docker.Cmd) > 1 {
+					args = task.Docker.Cmd[1:]
+				}
+			}
+			// get secrets
+			secrets, err := i.kube.CoreV1().Secrets(service.ProjectID).Get(secretName, metav1.GetOptions{})
+			if err != nil {
+				// error getting secrets!! return error!!
```
YourTechBud: Maybe we should return an error here

runner/utils/driver/istio/istio_helpers.go
```diff
@@ -381,6 +431,7 @@ func generateServiceAccount(service *model.Service) *v1.ServiceAccount {
 }
 
 func (i *Istio) generateDeployment(service *model.Service) *appsv1.Deployment {
+	preparedContainer, _ := i.prepareContainers(service)
```
YourTechBud: Shouldn't you be taking the volumes and setting it?

Issue #659, opened 2020-01-23
Can't query arrays in REST response from external service


lukastosic:
### Describe the bug

REST endpoint of an external service is responding with some data, among others there are fields that are array types. If I try to use fields from that array - I get null in the response

### Expected behaviour

I should be able to give graphql query to show only certain fields from the REST array response

### Actual behaviour 

Can only show entire field

### Steps to reproduce

Add some external service REST API that produces array.

In my sample I have response from REST API like:

```
{
    "longitude": "4.7346168",
    "latitude": "52.0894719",
    "fiberCoverage": [
        {
            "networkOwner": "KPN",
            "networkOperator": "KPN",
             # there are more fields
         },
        {
            "networkOwner": "TELE2",
            "networkOperator": "TELE2",
            "additionalInformationMessage": "4, Nearnet",
            # there are more fields
         }
}
```

I can use GraphQL query like:

```
{
  coverage(zipcode: "2411ND", number: "2") @service{
    latitude,
    longitude,
    fiberCoverage
  }  
}
```

And in GraphQL response - I would see entire content of the `fiberCoverage` field - complete array

but if I try to do something like this (I just want `networkOwner` name to be in the array)
```
{
  coverage(zipcode: "2411ND", number: "2") @service{
    latitude,
    longitude,
    fiberCoverage {
      networkOwner
    }
  }  
}
```

Then response from GraphQL looks like this:

```
{
  "coverage": {
    "latitude": "52.0894719",
    "longitude": "4.7346168",
    "fiberCoverage": {
      "networkOwner": null
    }
  }
}
```

### Your environment
- Space Cloud version: 0.15.3
- OS: Docker on widnows
- The client api version (if applicable):
- Browser (if applicable): Chrome

>If this bug restricts your use of space-cloud, give it a thumbs up reaction, so that we can determine which bugs need to be fixed immediately.
>👍

YourTechBud:
Ah! I see. We have never tested for this use case. Seems like an easy fix. Will slate it in for `v0.16.0`

Thanks for bringing this up @lukastosic

lukastosic:
@YourTechBud Thanks a lot!

YourTechBud:
Lets not close this till we have implemented it

lukastosic:
Sorry, wrong button :)

YourTechBud:
Coming in `v0.16.0`

PR #675, opened 2020-01-30
solved issue #659

sharadregoti:


gateway/utils/graphql/graphql.go
```diff
@@ -208,42 +208,15 @@ func (graph *Module) execGraphQLDocument(ctx context.Context, node ast.Node, tok
 
 		currentValue, err := utils.LoadValue(fmt.Sprintf("%s.%s", store["coreParentKey"], field.Name.Value), store)
 		if err != nil {
-			cb(nil, nil)
+			cb(nil, nil) // todo should we return error here
```
YourTechBud: This will be nil. Add a comment on top saying that if the field isn't found in the state means that field did not exist in the result. In that case, we simply return nil. However no error really exists so we do not return an error

PR #687, opened 2020-02-03
added command generate service

sharadregoti:


space-cli/actions.go
```diff
@@ -55,6 +56,16 @@ func actionBuildCode(c *cli.Context) error {
 	return nil
 }
 
+func actionGenerateService(c *cli.Context) error {
+	service, err := cmd.GenerateServiceConfigWithoutLogin()
+	if err != nil {
+		return err
+	}
+	data, _ := json.MarshalIndent(service, "", " ")
```
YourTechBud: Check the error here

space-cli/cmd/code.go
```diff
@@ -75,27 +73,26 @@ func generateServiceConfig(projects []*model.Projects, selectedaccount *model.Ac
...
 		Tasks: []model.Task{
 			{
 				ID:        serviceID,
 				Name:      serviceName,
-				Ports:     []model.Port{model.Port{Protocol: "http", Port: port}},
+				Ports:     []model.Port{{Name: "http", Protocol: "http", Port: port}},
 				Resources: model.Resources{CPU: 250, Memory: 512},
 				Docker:    model.Docker{Image: img},
-				Env:       map[string]string{"URL": selectedaccount.ServerUrl, "CMD": progCmd},
+				Env:       map[string]string{"URL": selectedaccount.ServerUrl, "CMD": progCmd}, // todo check why do we need this envs
```
YourTechBud: No need to set this anymore

space-cli/cmd/code.go
```diff
@@ -106,36 +103,97 @@ func RunDockerFile(s *model.ActionCode, loginResp *model.LoginResponse) error {
...
-				Source: dir,
-				Target: "/build",
-			},
-		}}, nil, "")
+	data, err := ioutil.ReadFile(fmt.Sprintf("%s/Dockerfile", dir))
 	if err != nil {
 		return err
 	}
-	if err := cli.ContainerStart(ctx, resp.ID, types.ContainerStartOptions{}); err != nil {
+	_, err = cli.ImageBuild(ctx, strings.NewReader(string(data)), types.ImageBuildOptions{})
```
YourTechBud: What will be the name of the image built?

space-cli/cmd/code.go
```diff
@@ -106,36 +103,97 @@ func RunDockerFile(s *model.ActionCode, loginResp *model.LoginResponse) error {
...
 	}
+	// todo store some name or for creating container
 	return nil
 }
+
+func GenerateServiceConfigWithoutLogin() (*model.Service, error) {
+	progLang, err := getProgLang()
+	if err != nil {
+		return nil, err
+	}
```
YourTechBud: We don't need this

space-cli/cmd/code.go
```diff
@@ -106,36 +103,97 @@ func RunDockerFile(s *model.ActionCode, loginResp *model.LoginResponse) error {
...
+	if dockerImage == "" {
+		dockerImage = fmt.Sprintf("%s/%s", projectNameID, serviceID)
+	}
+
+	dockerSecret := ""
+	if err := survey.AskOne(&survey.Input{Message: "Enter Docker Secret"}, &dockerSecret); err != nil {
+		return nil, err
+	}
```
YourTechBud: Only ask for this if it was not a private image. 
Let's do this. Keep the default value `none`. If the secret value is `none` then don't set the secret in the docker object. Also, make this a radio button kind of a thing. Let the user select from the list of available docker secrets

space-cli/cmd/code.go
```diff
@@ -106,36 +103,97 @@ func RunDockerFile(s *model.ActionCode, loginResp *model.LoginResponse) error {
...
+	secrets := []string{}
+	for _, value := range strings.Split(secret, ",") {
+		secrets = append(secrets, value)
+	}
+
+	var progCmd string
+	if err := survey.AskOne(&survey.Input{Message: "Enter Run Cmd", Default: strings.Join(getCmd(progLang), " ")}, &progCmd); err != nil {
+		return nil, err
+	}
```
YourTechBud: no need for this

space-cli/cmd/code.go
```diff
@@ -106,36 +103,97 @@ func RunDockerFile(s *model.ActionCode, loginResp *model.LoginResponse) error {
...
+
+	secret := ""
+	if err := survey.AskOne(&survey.Input{Message: "Enter File & Environment Secret (CSV)"}, &secret); err != nil {
+		return nil, err
+	}
+	secrets := []string{}
+	for _, value := range strings.Split(secret, ",") {
+		secrets = append(secrets, value)
+	}
```
YourTechBud: Not required. Allow the user to not enter anything. Using the default `none` option would be great here.
I would divide this into two questions - 1) ask if service has secrets - 2) If yes, show a multi select list of all secrets entered.

space-cli/cmd/code.go
```diff
@@ -106,36 +103,97 @@ func RunDockerFile(s *model.ActionCode, loginResp *model.LoginResponse) error {
...
+		Name:      serviceName,
+		ProjectID: projectNameID,
+		Version:   "v1",
+		Scale:     model.ScaleConfig{Replicas: 0, MinReplicas: 0, MaxReplicas: 100, Concurrency: 50},
+		Tasks: []model.Task{
+			{
+				ID:        serviceID,
+				Name:      serviceName,
+				Ports:     []model.Port{{Name: "http", Protocol: "http", Port: port}},
+				Resources: model.Resources{CPU: 250, Memory: 512},
```
YourTechBud: Maybe we should ask the user for number of resources as well?

space-cli/cmd/code.go
```diff
@@ -106,36 +103,97 @@ func RunDockerFile(s *model.ActionCode, loginResp *model.LoginResponse) error {
...
+				Docker:    model.Docker{Image: dockerImage, Secret: dockerSecret},
```
YourTechBud: Set the secret only if its not none. A simple way would be to set the `dockerSecret` value to `""` if its value was none

space-cli/actions.go
```diff
@@ -55,6 +56,16 @@ func actionBuildCode(c *cli.Context) error {
 	return nil
 }
 
+func actionGenerateService(c *cli.Context) error {
+	service, err := cmd.GenerateServiceConfigWithoutLogin()
```
YourTechBud: This function isn't returning the format we discussed

space-cli/actions.go
```diff
@@ -26,43 +32,71 @@ func setLogLevel(loglevel string) {
...
-		IsDeploy: true, //
-	}
-	if err := cmd.RunDockerFile(actionCodeStruct, loginResp); err != nil {
+	fileContent := new(model.GitOp)
+	if err := json.Unmarshal(data, fileContent); err != nil {
 		return err
 	}
-	return nil
-}
+	if fileContent.Type != "service" {
```
YourTechBud: Don't do this step

space-cli/actions.go
```diff
@@ -26,43 +32,71 @@ func setLogLevel(loglevel string) {
...
+		return fmt.Errorf("invalid type found in serivce file type should be (service)")
+	}
+	service := fileContent.Spec.(*model.Service)
```
YourTechBud: When you reading the file, spec will be parsed as a map string interface

space-cli/actions.go
```diff
@@ -26,43 +32,71 @@ func setLogLevel(loglevel string) {
 	}
 }
 
-func actionStartCode(c *cli.Context) error {
-	service, loginResp, err := cmd.CodeStart()
+func actionApply(c *cli.Context) error {
+	args := os.Args
+	if len(args) != 3 {
```
YourTechBud: I think you'll get just a single argument. Haven't you tested this?

space-cli/actions.go
```diff
@@ -26,43 +32,71 @@ func setLogLevel(loglevel string) {
...
 	if err != nil {
+		logrus.Error("error in apply service unable to marshal data - %v", err)
 		return err
 	}
-	actionCodeStruct := &model.ActionCode{
-		Service:  service,
-		IsDeploy: false,
-	}
-	if err := cmd.RunDockerFile(actionCodeStruct, loginResp); err != nil {
+	urlPath := strings.Replace(fileContent.Api, "{projectId}", service.ProjectID, 1)
```
YourTechBud: What if there is a `{key}` as well in the url? You need to make this dynamic my boy

space-cli/actions.go
```diff
@@ -26,43 +32,71 @@ func setLogLevel(loglevel string) {
...
+	resp, err := http.Post(fmt.Sprintf("http://localhost:4122%s", urlPath), "application/json", bytes.NewBuffer(requestBody))
```
YourTechBud: How did you get this url??. I would have taken it from server url field in the selected account

space-cli/actions.go
sharadregoti: image

YourTechBud: Cool

PR #723, opened 2020-02-11
keeping {project} in url and changing project-projectID,dbType-dbAlia…

prithvi081099:
…s and event-trigger to event/triggers


Jayesh333:
Have to close this as it has some bugs

PR #731, opened 2020-02-13
Data loader

sharadregoti:


gateway/modules/crud/dataloader.go
```diff
@@ -96,24 +96,24 @@ func newLoaderMap() *loaderMap {
...
+func (l *loaderMap) get(key string, m *Module) *dataloader.Loader {
 	l.lock.Lock()
 	defer l.lock.Unlock()
 
 	if _, ok := l.m[key]; !ok {
-		l.m[key] = graph.createLoader()
+		l.m[key] = m.createLoader()
 	}
 
 	return l.m[key]
```
YourTechBud: Use two functions here: `getLoader` and `createLoader`. `getLoader` will return a loader and a boolean indicating whether its present or not. Make getLoader use a `RLock` to increase parallelism. `createLoader` will use a `Lock` internally since its doing a mutation

gateway/modules/crud/dataloader.go
```diff
@@ -145,7 +145,8 @@ func (graph *Module) dataLoaderBatchFn(c context.Context, keys dataloader.Keys)
 				defer wg.Done()
 
 				// Execute the query
-				res, err := graph.crud.Read(ctx, req.DBType, graph.project, req.Col, &req.Req)
+				req.Req.Options.Prefix = ""
```
YourTechBud: Add a comment on why this line is so important. Also put a note to not remove this

gateway/modules/crud/dataloader.go
```diff
@@ -173,7 +174,9 @@ func (graph *Module) dataLoaderBatchFn(c context.Context, keys dataloader.Keys)
 	req := model.ReadRequest{Find: map[string]interface{}{"$or": holder.getWhereClauses()}, Operation: utils.All, Options: &model.ReadOptions{}}
 
 	// Fire the merged request
-	res, err := graph.crud.Read(ctx, dbType, graph.project, col, &req)
+
+	req.Options.Prefix = ""
```
YourTechBud: Add a comment on why this line is so important. Also put a note to not remove this

gateway/modules/crud/operations.go
```diff
@@ -54,7 +58,21 @@ func (m *Module) Read(ctx context.Context, dbAlias, project, col string, req *mo
...
 
-	n, result, err := crud.Read(ctx, project, col, req)
+	var n int64
+	var result interface{}
+	var dataLoader *dataloader.Loader
+	var key model.ReadRequestKey
+	if req.Options != nil {
+		if req.Options.Prefix != "" {
+			key = model.ReadRequestKey{DBType: dbAlias, Col: col, HasOptions: req.Options.HasOptions, Req: *req}
+			dataLoader = newLoaderMap().get(fmt.Sprintf("%s-%s", req.Options.Prefix, uuid.New().String()), m)
```
YourTechBud: A few problems here
- If you are creating a newLoaderMap for each request you will never batch the read operations. Have a single globally defined loadermap
- If you add a unique uuid for every key here, you'll never batch the read operations, since each read operations will have a different data loader struct.

gateway/modules/crud/operations.go
```diff
@@ -54,7 +58,21 @@ func (m *Module) Read(ctx context.Context, dbAlias, project, col string, req *mo
 		return nil, err
 	}
 
-	n, result, err := crud.Read(ctx, project, col, req)
+	var n int64
+	var result interface{}
+	var dataLoader *dataloader.Loader
```
YourTechBud: This should be initialised in the `Init` function of the module

gateway/modules/crud/operations.go
```diff
@@ -54,7 +58,21 @@ func (m *Module) Read(ctx context.Context, dbAlias, project, col string, req *mo
...
+	var key model.ReadRequestKey
+	if req.Options != nil {
+		if req.Options.Prefix != "" {
+			key = model.ReadRequestKey{DBType: dbAlias, Col: col, HasOptions: req.Options.HasOptions, Req: *req}
+			dataLoader = newLoaderMap().get(fmt.Sprintf("%s-%s", req.Options.Prefix, uuid.New().String()), m)
+			result, err = dataLoader.Load(ctx, key)()
+			// clear the key
+			_ = dataLoader.Clear(ctx, key)
```
YourTechBud: You shouldn't clear the key here. The user of the crud module will call clear this key.

gateway/modules/crud/operations.go
```diff
@@ -54,7 +58,21 @@ func (m *Module) Read(ctx context.Context, dbAlias, project, col string, req *mo
...
-	n, result, err := crud.Read(ctx, project, col, req)
+	var n int64
+	var result interface{}
+	var dataLoader *dataloader.Loader
+	var key model.ReadRequestKey
+	if req.Options != nil {
+		if req.Options.Prefix != "" {
+			key = model.ReadRequestKey{DBType: dbAlias, Col: col, HasOptions: req.Options.HasOptions, Req: *req}
+			dataLoader = newLoaderMap().get(fmt.Sprintf("%s-%s", req.Options.Prefix, uuid.New().String()), m)
+			result, err = dataLoader.Load(ctx, key)()
```
YourTechBud: We need the `n` variable or else the metrics module will go for a toss

gateway/utils/graphql/read.go
```diff
@@ -12,27 +12,27 @@ import (
...
 	}
 
-	dataLoader := loader.get(getFieldName(field)+"."+store["path"].(string)+".linked."+col, graph)
+	// dataLoader := loader.get(getFieldName(field)+"."+store["path"].(string)+".linked."+col, graph)
 
 	go func() {
-		// Create dataloader key
-		key := model.ReadRequestKey{DBType: dbType, Col: col, HasOptions: false, Req: *req}
-		result, err := dataLoader.Load(ctx, key)()
+		req.Options.Prefix = fmt.Sprintf("%s.%s.linked.%s", getFieldName(field), store["path"].(string), col)
```
YourTechBud: Prepend a `ksuid` to the prefix you are generating here. 

**You will not generate the ksuid here or it will beat the purpose**.

The `ksuid` will be generated when the request is being made and will be passed down till here. **Don't store the KSUID in the graph module struct**. The validity of a KSUID should be a for a single request only

gateway/utils/graphql/read.go
```diff
@@ -57,12 +57,12 @@ func (graph *Module) execReadRequest(ctx context.Context, field *ast.Field, toke
...
 	}
 
-	dataLoader := loader.get(getFieldName(field)+"."+store["path"].(string), graph)
+	// dataLoader := loader.get(getFieldName(field)+"."+store["path"].(string), graph)
 
 	go func() {
-		// Create dataloader key
-		key := model.ReadRequestKey{DBType: dbType, Col: col, HasOptions: hasOptions, Req: *req}
-		result, err := dataLoader.Load(ctx, key)()
+		req.Options.Prefix = fmt.Sprintf("%s.%s", getFieldName(field), store["path"].(string))
```
YourTechBud: Same as my previous review

gateway/modules/crud/dataloader.go
```diff
@@ -87,33 +87,22 @@ func (holder *resultsHolder) fillErrorMessage(err error) {
 	holder.Unlock()
 }
 
-type loaderMap struct {
-	lock sync.Mutex
-	m    map[string]*dataloader.Loader
+func (m *Module) getLoader(key string) (*dataloader.Loader, bool) {
+	loader, ok := m.dataLoader.loaderMap[key]
```
YourTechBud: Add comment that rlock is not required

gateway/modules/crud/operations.go
```diff
@@ -53,7 +54,15 @@ func (m *Module) Read(ctx context.Context, dbAlias, project, col string, req *mo
...
 		return nil, err
 	}
-
+	if req.IsBatch {
+		key := model.ReadRequestKey{DBType: dbAlias, Col: col, HasOptions: req.Options.HasOptions, Req: *req}
+		dataLoader, ok := m.getLoader(fmt.Sprintf("%s-%s-%s", project, dbAlias, col))
+		if !ok {
+			dataLoader = m.createLoader(fmt.Sprintf("%s-%s-%s", project, dbAlias, col))
+		}
+		defer dataLoader.Clear(ctx, key) // clear the key
```
YourTechBud: We shouldn't do this here. Create a new loader in set config. No need to clear the keys at all

PR #739, opened 2020-02-15
ignored metrics error and mounted host file to service containers

Jayesh333:


runner/utils/driver/docker/helpers.go
```diff
@@ -62,3 +63,22 @@ func (d *docker) pullImageIfDoesntExists(ctx context.Context, projectId string,
 	}
 	return nil
 }
+
+func getSpaceCloudHostsFilePath() string {
```
YourTechBud: You cannot use these functions here since these will be invoked inside the runner and not on the host

YourTechBud: Instead accept the `HOSTS_FILE_PATH` environment variable like we are doing for secrets in the `NewDockerDriver` method. Don't forget to add the env variable in space cli setup

PR #748, opened 2020-02-15
implemented delete collections method for embedded db

KavishShah09:


gateway/modules/crud/bolt/collections.go
```diff
@@ -14,5 +16,21 @@ func (b *Bolt) GetCollections(ctx context.Context, project string) ([]utils.Data
 
 // DeleteCollection deletes collection / tables name of specified database
 func (b *Bolt) DeleteCollection(ctx context.Context, project, col string) error {
-	return fmt.Errorf("error deleting collection operation not supported for selected database")
+	err := b.client.View(func(tx *bbolt.Tx) error {
```
sharadregoti: b.client.View is only used for read transactions
as you will performing a delete operation 
you should use b.client.Update for read write transactions
refer delete.go of bboltdb

gateway/modules/crud/bolt/collections.go
```diff
@@ -14,5 +16,21 @@ func (b *Bolt) GetCollections(ctx context.Context, project string) ([]utils.Data
...
+		b := tx.Bucket([]byte(project))
```
sharadregoti: check b == nil
refer read.go of boltdb

sharadregoti:
write tests for the above functions

Issue #706, opened 2020-02-08
Auto batch inserts optionally based on key


sharadregoti:
>If you want this feature to be implemented, give it a thumbs up reaction, so that we can determine which features are important to you.
>👍

YourTechBud:
Coming in `v0.17.0`

PR #775, opened 2020-02-20
Batch inserts

sharadregoti:


YourTechBud:
Can we add a max batchee items limit as well?

sharadregoti:
you want to say
number of items to be batched in a time frame am i correct ?

YourTechBud:
Yes. So in high throughput situations, the insert will happen before the timer if the max limit is reached. This will happen on a per collection basis

sharadregoti:
Got it

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+)
+
+// CloseBatchOperation closes all go routines associated with individual collection for batch operation
+func (m *Module) CloseBatchOperation() {
+	m.Lock()
+	defer m.Unlock()
+	for _, value := range m.batchMapTableToChan {
+		for _, channels := range value {
+			close(channels.close)
```
YourTechBud: You should be writing into this close channel

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,98 @@
...
+
+// data structure for storing a group of channels
+// against particular table/collection of particular database
+type (
+	batchMap map[string]map[string]batchChannels // key here is dbAlias & it's value's key is table name
+
+	batchChannels struct {
+		request  batchRequestChan
+		response chan error
+		close    chan bool
```
YourTechBud: make this a channel of struct

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	m.batchMapTableToChan = batch
+	// log.Println("go routines after creating executors -", runtime.NumGoroutine())
+}
+
+func (m *Module) insertBatchExecutor(done chan bool, response chan error, addInsertToBatchCh batchRequestChan, batchTime int, dbAlias, tableName string) {
+	project := ""
+	batchRequests := make([]interface{}, 0)
+	if batchTime <= 0 { // when new project is created set default time to 200 milli seconds
+		batchTime = 200
```
YourTechBud: default is 100 millis

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	ticker := time.NewTicker(time.Duration(batchTime) * time.Millisecond)
+	for {
+		select {
+		case <-done:
+			ticker.Stop()
+			logrus.Debugf("closing batcher for database %s table %s", dbAlias, tableName)
+			return
```
YourTechBud: close all the channels here

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,98 @@
...
+		case v := <-addInsertToBatchCh:
+			project = v.project
```
YourTechBud: what if you get requests for the same collection present in two different projects?

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,98 @@
...
+			batchRequests = append(batchRequests, v.document.([]interface{})...)
```
YourTechBud: If the document is always going to be an array, make it type `[]interface{}`

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,98 @@
...
+			log.Println("added batch request", batchRequests)
+		case <-ticker.C:
+			if len(batchRequests) != 0 {
+				ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second) // TODO CHECK CONTEXT TIME
```
YourTechBud: make it 30 seconds

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/spaceuptech/space-cloud/gateway/utils"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spaceuptech/space-cloud/gateway/config"
+)
+
+// data structure for storing a group of channels
+// against particular table/collection of particular database
+type (
+	batchMap map[string]map[string]batchChannels // key here is dbAlias & it's value's key is table name
```
YourTechBud: maybe you shoukd add one more depth to accomodate project

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,98 @@
...
+	for {
+		select {
+		case <-done:
+			ticker.Stop()
+			logrus.Debugf("closing batcher for database %s table %s", dbAlias, tableName)
+			return
+		case v := <-addInsertToBatchCh:
+			project = v.project
+			batchRequests = append(batchRequests, v.document.([]interface{})...)
+			log.Println("added batch request", batchRequests)
```
YourTechBud: Make it a `logrus.Debug`. Add some more context here like `project`, `dbAlias`, `collection` and length of array instead of printing the array straight up

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,98 @@
...
+		case <-ticker.C:
+			if len(batchRequests) != 0 {
+				ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second) // TODO CHECK CONTEXT TIME
+				if err := m.Create(ctx, dbAlias, project, tableName, &model.CreateRequest{Operation: utils.All, Document: batchRequests}); err != nil {
+					logrus.Errorf("error executing batch request for database %s table %s - %s", dbAlias, tableName, err)
+					response <- err
+				}
+				cancel()                               // close context
+				batchRequests = make([]interface{}, 0) // clear the requests array
+				response <- nil
```
YourTechBud: One response going to everyone? That wont work. You to make a channel per request and send response to each one.

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,98 @@
...
+		case v := <-addInsertToBatchCh:
+			project = v.project
+			batchRequests = append(batchRequests, v.document.([]interface{})...)
+			log.Println("added batch request", batchRequests)
+		case <-ticker.C:
+			if len(batchRequests) != 0 {
+				ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second) // TODO CHECK CONTEXT TIME
+				if err := m.Create(ctx, dbAlias, project, tableName, &model.CreateRequest{Operation: utils.All, Document: batchRequests}); err != nil {
+					logrus.Errorf("error executing batch request for database %s table %s - %s", dbAlias, tableName, err)
+					response <- err
```
YourTechBud: One response going to everyone? That wont work. You to make a channel per request and send response to each one.
You forgot to `cancel()` the context and break

gateway/modules/crud/operations.go
```diff
@@ -21,6 +24,18 @@ func (m *Module) Create(ctx context.Context, dbAlias, project, col string, req *
...
 
+	// add the request for batch operation
+	if req.IsBatch {
+		ch, ok := m.batchMapTableToChan[dbAlias][col] // get channel for specified table
+		if !ok {
+			logrus.Errorf("error converting insert request to batch request unable to find channel for database %s & collection %s", dbAlias, col)
+			return fmt.Errorf("cannot find channel for database %s & collection %s", dbAlias, col)
+		}
+		// TODO VALIDATION OF PROJECT
+		ch.request <- batchRequest{document: req.Document, project: project}
```
YourTechBud: Create a new channel and pass that as well here

gateway/modules/crud/operations.go
```diff
@@ -21,6 +24,18 @@ func (m *Module) Create(ctx context.Context, dbAlias, project, col string, req *
...
+		return <-ch.response
```
YourTechBud: You'll get the response in the new channel you just created. Don't forget to close the channel in differ

gateway/utils/syncman/operations.go
```diff
@@ -156,6 +156,7 @@ func (s *Manager) SetProjectConfig(ctx context.Context, project *config.Project)
 }
 
 func (s *Manager) setProject(ctx context.Context, project *config.Project) error {
+	s.crud.CloseBatchOperation()
```
YourTechBud: You should call this function in the  `SetConfig` of the crud module

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+		if dbInfo.Enabled {
+			for tableName, tableInfo := range dbInfo.Collections {
+				done := make(chan struct{})                                          // channel for closing go routine
+				addInsertToBatchCh := make(batchRequestChan, tableInfo.BatchRecords) // channel for adding request to batch op
+				go m.insertBatchExecutor(done, addInsertToBatchCh, dbInfo.BatchTime, project, dbAlias, tableName, tableInfo.BatchRecords)
+				if batch[project] == nil {
+					batch[project] = map[string]map[string]batchChannels{dbAlias: {tableName: {request: addInsertToBatchCh, close: done}}}
+					continue
+				}
+				batch[project][dbAlias][tableName] = batchChannels{request: addInsertToBatchCh, close: done}
```
YourTechBud: What if `batch[project][dbAlias]` is `nil`? You should check for that as well.

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+// initBatchOperation creates go routines for executing batch operation associated with individual collection
+func (m *Module) initBatchOperation(project string, crud config.Crud) {
+	batch := batchMap{}
+	for dbAlias, dbInfo := range crud {
+		if dbInfo.Enabled {
+			for tableName, tableInfo := range dbInfo.Collections {
+				done := make(chan struct{})                                          // channel for closing go routine
+				addInsertToBatchCh := make(batchRequestChan, tableInfo.BatchRecords) // channel for adding request to batch op
```
YourTechBud:
```suggestion
				addInsertToBatchCh := make(batchRequestChan, 20) // channel for adding request to batch op
```

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+// initBatchOperation creates go routines for executing batch operation associated with individual collection
+func (m *Module) initBatchOperation(project string, crud config.Crud) {
+	batch := batchMap{}
+	for dbAlias, dbInfo := range crud {
+		if dbInfo.Enabled {
+			for tableName, tableInfo := range dbInfo.Collections {
+				done := make(chan struct{})                                          // channel for closing go routine
```
YourTechBud: Whu not call this `closeC`

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+// data structure for storing a group of channels
+// against particular table/collection of particular database
+type (
+	batchMap map[string]map[string]map[string]batchChannels // keys are project name, dbAlias, table name
+
+	batchChannels struct {
+		request batchRequestChan
+		close   chan struct{}
```
YourTechBud: rename this to `closeC`

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+	m.batchMapTableToChan = batch
+}
+
+func (m *Module) insertBatchExecutor(done chan struct{}, addInsertToBatchCh batchRequestChan, batchTime int, project, dbAlias, tableName string, batchRecordLimit int) {
+	noOfRequests := 0
+	responseChannels := make([]chan error, 0)
+	batchRequests := make([]interface{}, 0)
+	if batchTime <= 0 { // when new project is created set default time to 200 milli seconds
```
YourTechBud:
```suggestion
	if batchTime <= 0 { // when new project is created set default time to 100 milli seconds
```

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	if batchRecordLimit <= 0 {
+		batchRecordLimit = 100 // when new project is created set default batch record limit to 100
+	}
+	ticker := time.NewTicker(time.Duration(batchTime) * time.Millisecond)
+	for {
+		select {
+		case <-done:
+			ticker.Stop()
+			close(addInsertToBatchCh)
```
YourTechBud: Write a comment here that this is a safe operation since `SetConfig` will hold a lock preventing others from writing into this channel after its closed

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+			close(done)
+			logrus.Debugf("closing batcher for database %s table %s", dbAlias, tableName)
+			return
+		case v := <-addInsertToBatchCh:
+			noOfRequests++
```
YourTechBud: Not a right place to do this.

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+			responseChannels = append(responseChannels, v.response)
+			switch docType := v.document.(type) {
+			case map[string]interface{}:
+				batchRequests = append(batchRequests, docType)
```
YourTechBud: Add `noOfRequests++` here

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+			case []interface{}:
+				batchRequests = append(batchRequests, docType...)
```
YourTechBud: Add `noOfequests += len(docType)`

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+			}
+			if noOfRequests == batchRecordLimit {
```
YourTechBud: Do we really need `noOfRequests` at all? Why not use the length of array

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+				m.executeBatch(project, dbAlias, tableName, batchRequests, responseChannels)
+				batchRequests = make([]interface{}, 0) // clear the requests array
+				noOfRequests = 0
+				responseChannels = make([]chan error, 0) // clear the response channels array
+				// reset ticker
+				ticker.Stop()
+				ticker = time.NewTicker(time.Duration(batchTime) * time.Millisecond)
+			}
+		case <-ticker.C:
+			if len(batchRequests) != 0 {
```
YourTechBud:
```suggestion
			if len(batchRequests) > 0 {
```

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func (m *Module) executeBatch(project, dbAlias, tableName string, batchRequests []interface{}, responseChannels []chan error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	if err := m.Create(ctx, dbAlias, project, tableName, &model.CreateRequest{Operation: utils.All, Document: batchRequests}); err != nil {
+		logrus.Errorf("error executing batch request for database %s table %s - %s", dbAlias, tableName, err)
+		for _, responseChan := range responseChannels {
+			responseChan <- err
+		}
```
YourTechBud: No need to write this twice. Make the code DRY

gateway/modules/crud/crud.go
```diff
@@ -87,6 +90,7 @@ func (m *Module) getCrudBlock(dbType string) (Crud, error) {
 func (m *Module) SetConfig(project string, crud config.Crud) error {
 	m.Lock()
 	defer m.Unlock()
+	m.CloseBatchOperation()
```
YourTechBud: No need to export this

gateway/modules/crud/operations.go
```diff
@@ -21,6 +24,20 @@ func (m *Module) Create(ctx context.Context, dbAlias, project, col string, req *
...
+		defer close(response)
+		ch, ok := m.batchMapTableToChan[project][dbAlias][col] // get channel for specified table
+		if !ok {
+			logrus.Errorf("error converting insert request to batch request unable to find channel for database %s & collection %s", dbAlias, col)
+			return fmt.Errorf("cannot find channel for database %s & collection %s", dbAlias, col)
+		}
+		// TODO VALIDATION OF PROJECT
+		ch.request <- batchRequest{document: req.Document, response: response}
+		return <-response
+	}
```
YourTechBud: This cannot bypass the hooks (`m.hook.Create`, `m.metricHook`, `m.hooks.Stage`)

gateway/modules/crud/batcher.go
sharadregoti: this condition will not occur as if a project doen't exists
the whole structure gets initialized for that project
next time for the same project the whole structure is already initialized
see the if statement above

gateway/modules/crud/batcher.go
sharadregoti: but the channel length should be equal to the length of records to be batched

YourTechBud: No. It has nothing to do with the batch size

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,129 @@
...
+		if dbInfo.Enabled {
+			for tableName, tableInfo := range dbInfo.Collections {
+				closeC := make(chan struct{})                    // channel for closing go routine
+				addInsertToBatchCh := make(batchRequestChan, 20) // channel for adding request to batch op
+				go m.insertBatchExecutor(closeC, addInsertToBatchCh, dbInfo.BatchTime, project, dbAlias, tableName, tableInfo.BatchRecords)
+				if batch[project] == nil {
+					batch[project] = map[string]map[string]batchChannels{dbAlias: {tableName: {request: addInsertToBatchCh, closeC: closeC}}}
+					continue
+				}
+				batch[project][dbAlias][tableName] = batchChannels{request: addInsertToBatchCh, closeC: closeC}
```
YourTechBud: What if `batch[project][dbAlias] id nil or that keg doesn't exist? You should check for that as well

gateway/modules/crud/batcher.go
```diff
@@ -0,0 +1,129 @@
...
+func (m *Module) executeBatch(project, dbAlias, tableName string, batchRequests []interface{}, responseChannels []batchResponseChan) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	if err := m.InternalCreate(ctx, dbAlias, project, tableName, &model.CreateRequest{Operation: utils.All, Document: batchRequests}); err != nil {
+		logrus.Errorf("error executing batch request for database %s table %s - %s", dbAlias, tableName, err)
+		m.sendResponses(responseChannels, batchResponse{err: err, docsInserted: 0})
+		return
+	}
+	// send response to all client request
+	m.sendResponses(responseChannels, batchResponse{err: nil, docsInserted: int64(len(batchRequests))})
```
YourTechBud: The docs inserted you've put here is incorrect since this is the batch size not the number of documents inserted by individual requests

YourTechBud: You should simply send the err which is nil

gateway/modules/crud/helpers.go
```diff
@@ -0,0 +1,20 @@
...
+	response := make(batchResponseChan, 1)
+	defer close(response)
+	ch, ok := m.batchMapTableToChan[project][dbAlias][col] // get channel for specified table
+	if !ok {
+		logrus.Errorf("error converting insert request to batch request unable to find channel for database %s & collection %s", dbAlias, col)
+		return 0, fmt.Errorf("cannot find channel for database %s & collection %s", dbAlias, col)
+	}
+	ch.request <- batchRequest{document: doc, response: response}
+	result := <-response
+	return result.docsInserted, result.err
```
YourTechBud: Instead of using the number of docs returned from the channel, check the length of the docs interface. If its a map the number is one. If its an array, simply check for the length of the array

gateway/modules/crud/operations.go
```diff
@@ -27,8 +27,14 @@ func (m *Module) Create(ctx context.Context, dbAlias, project, col string, req *
 		return err
 	}
 
-	// Perform the create operation
-	n, err := crud.Create(ctx, project, col, req)
+	var n int64
+	// add the request for batch operation
```
YourTechBud: Comment should go inside if statement

gateway/modules/crud/batcher.go
sharadregoti: this condition will not occur
as the map 's value for every project will be nil
and i check for nil, if nill i initialze the complete remaining data structure
check the above if case

gateway/modules/crud/batcher.go
sharadregoti: @YourTechBud but for metrics module we need count of docs inserted
and the batchRequests variable represents the number of recordes inserted 
why can't it represent the count ?

sharadregoti: ok i understood from the review given below

YourTechBud:
Solves #706

Issue #771, opened 2020-02-19
The routing module doesn't relay back the headers of the response.


YourTechBud:
>If this bug restricts your use of space-cloud, give it a thumbs up reaction, so that we can determine which bugs need to be fixed immediately.
>👍

YourTechBud:
Fixed in `v0.17.0`

PR #783, opened 2020-02-24
added support for weighted routing in the routing module

YourTechBud:


YourTechBud:
Fixes #783 
Fixes #771

PR #797, opened 2020-02-29
1)solved list bug problem in gcp

sharadregoti:
2)now updating security rules for existing rule doesn't give error
3)add method of checking connection state of storage module


gateway/modules/filestore/operations.go
```diff
@@ -170,3 +170,16 @@ func (m *Module) DoesExists(ctx context.Context, project, token, path string) er
...
 	return m.store.DoesExists(path)
 }
+
+// GetState checks if selected storage is active
+func (m *Module) GetState(ctx context.Context) error {
+	// Exit if file storage is not enabled
+	if !m.IsEnabled() {
+		return errors.New("This feature isn't enabled")
+	}
+	// TODO BYPASSING SECURITY RULE CHECK
```
YourTechBud: Remove this todo

gateway/modules/filestore/operations.go
```diff
@@ -170,3 +170,16 @@ func (m *Module) DoesExists(ctx context.Context, project, token, path string) er
...
+	m.RLock()
+	defer m.RUnlock()
+	// Read the state from file storage
+	return m.store.GetState()
```
YourTechBud: Pass the context here

gateway/modules/filestore/gcpstorage/raw.go
```diff
@@ -11,3 +12,11 @@ func (g *GCPStorage) DoesExists(path string) error {
 	}
 	return nil
 }
+
+// GetState checks if sc is able to query gcp storage
+func (g *GCPStorage) GetState() error {
+	if _, err := g.client.Bucket(g.bucket).Object("/").Attrs(context.TODO()); err != nil && err != storage.ErrObjectNotExist {
```
YourTechBud: Use the context from the parameter

PR #833, opened 2020-03-11
Modules

KavishShah09:


gateway/utils/syncman/syncman.go
```diff
@@ -32,6 +35,11 @@ type Manager struct {
 
 	// For authentication
 	adminMan *admin.Manager
+
+	// Modules
```
YourTechBud: Remove cb from the struct

gateway/utils/server/server.go
```diff
@@ -185,47 +142,47 @@ func (s *Server) LoadConfig(config *config.Config) error {
 
 		// Always set the config of the crud module first
 		// Set the configuration for the crud module
-		if err := s.crud.SetConfig(p.ID, p.Modules.Crud); err != nil {
```
YourTechBud: Remove the `LoadConfig` method. We don't need it anymore

gateway/utils/syncman/syncman.go
KavishShah09: Some functions still call s.cb so should I call SetProjectConfig instead at those places?

YourTechBud: Which functions?

PR #838, opened 2020-03-12
Schema module fixes

### LINKED ISSUE #581: Cannot change field type of indexed field
when changing the Fieldtype, column is drop first and then index is dropped

### LINKED ISSUE #751: Throw error on specifying unknown directives in the db schema
Space Cloud does not throw an error even if we specify an unknown directive in the schema. 

For example, I specified a schema which had `createAt` directive instead of `createdAt` by mistake. I could only find out my mistake while querying the corresponding table. Instead, if space cloud could provide error while schema creation itself, development and debugging time could be saved by the developer

### LINKED ISSUE #828: changing schema throws error
### Describe the bug
when schema created
type users { 
	email: String!     
	id: ID! @primary    
	name: String     
	uid: String!     
}
changing the above schema to show below gives error 
"error":"pq: multiple primary keys for table \"users\" are not allowed"
type users { 
	email: ID! @primary
	name: String     
	uid: String!     
}
### Your environment
- Space Cloud version:v0.17.0
- OS:linux

### LINKED ISSUE #830: changing column name of primary key throws error in schema
### Describe the bug
current schema
type clusters {
  id: ID! @primary
  project_id : String!
  url : String!
}
scheam after changing name of primary key column
type clusters {
  cluster_id: ID! @primary
  project_id : String!
  url : String!
}
throws error
{"error":"pq: column \"cluster_id\" contains null values"}

- Space Cloud version: v0.17.0
- OS:
- The client api version (if applicable):
- Browser (if applicable):

>👍

### LINKED ISSUE #831: removing primary key from schema gives error for postgres
### Describe the bug
current shema
type clusters {
  cluster_id: ID! @primary
  project_id : ID
	user_id : ID!
	type : ID!
  url : String!
}
scheam after removing primary key
type clusters {
  cluster_id: ID!
  project_id : ID
	user_id : ID!
	type : ID!
  url : String!
}
the table is empty
got error
{"error":"pq: constraint \"c_clusters_cluster_id\" of relation \"clusters\" does not exist"}
### Your environment
- Space Cloud version: v0.17.0
- OS:
- The client api version (if applicable):
- Browser (if applicable):

>👍

YourTechBud:
Fixes a few usability bugs in the schema module
- Fixes #581 
- Fixes #751
- Fixes #828 
- Fixes #830 
- Fixes #831

### DIFF
--- a/gateway/modules/schema/helpers.go
+++ b/gateway/modules/schema/helpers.go
@@ -69,7 +69,7 @@ func (c *creationModule) addNotNull() string {
 		return ""
 	}
 
-	c.currentColumnInfo.IsFieldTypeRequired = true
+	c.currentColumnInfo.IsFieldTypeRequired = true // Mark the field as processed
 	switch utils.DBType(dbType) {
 	case utils.MySQL:
 		return "ALTER TABLE " + getTableName(c.project, c.TableName, c.removeProjectScope) + " MODIFY " + c.ColumnName + " " + c.columnType + " NOT NULL"
@@ -130,7 +130,7 @@ func (c *creationModule) addPrimaryKey() string {
 		return ""
 	}
 
-	c.currentColumnInfo.IsPrimary = true
+	c.currentColumnInfo.IsPrimary = true // Mark the field as processed
 	switch utils.DBType(dbType) {
 	case utils.MySQL:
 		return "ALTER TABLE " + getTableName(c.project, c.TableName, c.removeProjectScope) + " ADD PRIMARY KEY (" + c.ColumnName + ")"
@@ -161,7 +161,7 @@ func (c *creationModule) removePrimaryKey() string {
 }
 
 func (c *creationModule) addForeignKey() string {
-	c.currentColumnInfo.IsForeign = true
+	c.currentColumnInfo.IsForeign = true // Mark the field as processed
 	return "ALTER TABLE " + getTableName(c.project, c.TableName, c.removeProjectScope) + " ADD CONSTRAINT c_" + c.TableName + "_" + c.ColumnName + " FOREIGN KEY (" + c.ColumnName + ") REFERENCES " + getTableName(c.project, c.realColumnInfo.JointTable.Table, c.removeProjectScope) + " (" + c.realColumnInfo.JointTable.To + ")"
 }
 
@@ -193,7 +193,7 @@ func (c *creationModule) addDefaultKey() string {
 		return ""
 	}
 
-	c.currentColumnInfo.IsDefault = true
+	c.currentColumnInfo.IsDefault = true // Mark the field as processed
 	c.currentColumnInfo.Default = c.realColumnInfo.Default
 	switch utils.DBType(dbType) {
 	case utils.MySQL:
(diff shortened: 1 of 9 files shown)


gateway/modules/schema/helpers.go
```diff
@@ -401,62 +440,49 @@ type indexStruct struct {
...
+			// Set the default sorting scheme if not present
+			if columnInfo.IndexInfo.Sort == "" {
+				columnInfo.IndexInfo.Sort = "asc"
 			}
-			if currentColumnInfo.IsUnique {
-				currentIndexMap[currentColumnInfo.IndexInfo.Group].IsIndexUnique = true
+
+			// Make sure the sort type is either ascending or descending
+			if !(columnInfo.IndexInfo.Sort == "asc" || columnInfo.IndexInfo.Sort == "desc") {
+				return nil, fmt.Errorf("invalid sort type (%s) provided for field (%s)", columnInfo.IndexInfo.Sort, columnInfo.FieldName)
```
sharadregoti: can we do this in the parse function of schema module ?

gateway/utils/handlers/graphql.go
```diff
@@ -29,6 +29,10 @@ func HandleGraphQLRequest(graphql *graphql.Module, syncMan *syncman.Manager) htt
 			return
 		}
 
+		if projectConfig.ContextTime == 0 {
```
sharadregoti: can we set the default value on creation of project ?

gateway/modules/schema/helpers.go
```diff
@@ -352,17 +365,43 @@ func (c *creationModule) modifyColumn() []string {
...
 
-// modifyColumnType drop the column then creates a new column with provided type
-func (c *creationModule) modifyColumnType(dbType string) []string {
-	queries := []string{}
+func (c *creationModule) removeDirectives() []string {
+	var queries []string
 
 	if c.currentColumnInfo.IsForeign {
 		queries = append(queries, c.removeForeignKey()...)
+		c.currentColumnInfo.IsForeign = false
```
sharadregoti: specify a comment why this is important

gateway/modules/schema/helpers.go
```diff
@@ -128,6 +130,7 @@ func (c *creationModule) addPrimaryKey() string {
 		return ""
 	}
 
+	c.currentColumnInfo.IsPrimary = true
```
sharadregoti: specify why this is important

gateway/utils/handlers/graphql.go
YourTechBud: What if the project already existed? As in space cloud is using config of `v0.16.0`

gateway/modules/schema/helpers.go
YourTechBud: I don't know. This was already here. I'll move it somewhere within SetConfig

gateway/modules/schema/helpers.go
YourTechBud: We just removed the foreign key. I think its pretty straight forward why i did that. Was to mark foreign key as removed to prevent this from being processed again

gateway/modules/schema/helpers.go
YourTechBud: Not needed

PR #849, opened 2020-03-15
added header to handlers

prithvi081099:


runner/server/handle_secrets.go
```diff
@@ -223,9 +259,12 @@ func (s *Server) handleDeleteSecretKey() http.HandlerFunc {
 		// setSecretKey
 		if err := s.driver.DeleteKey(projectID, name, key); err != nil {
 			logrus.Errorf("Failed to list secret - %s", err.Error())
+			w.Header().Set("Content-Type", "application/json")
 			utils.SendErrorResponse(w, r, http.StatusInternalServerError, err)
 			return
 		}
+
+		w.Header().Set("Content-Type", "application/json")
```
YourTechBud: Instead of setting header here, set it inside `utils.SendEmptySuccessResponse`. Do this everywhere.

runner/server/handle_secrets.go
```diff
@@ -223,9 +259,12 @@ func (s *Server) handleDeleteSecretKey() http.HandlerFunc {
 		// setSecretKey
 		if err := s.driver.DeleteKey(projectID, name, key); err != nil {
 			logrus.Errorf("Failed to list secret - %s", err.Error())
+			w.Header().Set("Content-Type", "application/json")
```
YourTechBud: Instead of setting error here. Set it inside `utils.SendErrorResponse`. Do this everywhere

PR #857, opened 2020-03-16
mongo extra entry

rushikesh-kulkarni:


gateway/modules/schema/validatecreate.go
```diff
@@ -16,6 +16,20 @@ import (
 // SchemaValidator function validates the schema which it gets from module
 func (s *Schema) SchemaValidator(col string, collectionFields model.Fields, doc map[string]interface{}) (map[string]interface{}, error) {
 
+	for schemaKey := range doc {
+		i := 1
+		for fieldKey := range collectionFields {
```
YourTechBud: Collection fields is a map. You don't need to loop here. You can simply do something like this.
```golang
if _, p := collectionFields[schemaKey]; !p {
  return nil, errors.New("The field " + wrongMongoFieldName + " is not present in schema of " + col)
}
```
You don't event need to play with the index

gateway/modules/schema/validatecreate.go
```diff
@@ -60,7 +74,7 @@ func (s *Schema) SchemaValidator(col string, collectionFields model.Fields, doc
 	return mutatedDoc, nil
 }
 
-// ValidateCreateOperation validates schema on create operation
+// ValidateCreateOperation validates  schema on create operation
```
YourTechBud:
```suggestion
// ValidateCreateOperation validates schema on create operation
```

PR #893, opened 2020-03-30
Changed structure of metrics collection in gateway & runner

sharadregoti:


gateway/utils/metrics/metrics.go
```diff
@@ -1,69 +1,76 @@
...
+	if config.IsDisabled {
 		return new(Module), nil
 	}
 
 	// Initialise the sink
-	c, err := initialiseSink(config)
-	if err != nil {
-		return nil, err
-	}
+	conn := api.New("projectmetrics", "localhost:4122", false).DB("postgres")
```
YourTechBud, line 49: Project name is `spacecloud` and db name is `db`

gateway/utils/metrics/metrics.go
```diff
@@ -1,69 +1,76 @@
...
 
 	// Create a new metrics module
-	m := &Module{nodeID: nodeID, sink: c, config: *config}
-
+	m := &Module{nodeID: nodeID, clusterID: clusterID, sink: conn, config: *config, syncMan: syncMan, isProd: isProd, eventing: map[string]int{}}
+	logrus.Println("metrics started")
```
YourTechBud, line 53: Don't log this

gateway/utils/metrics/operations.go
```diff
@@ -7,36 +7,37 @@ import (
...
 
-// AddEgress add the bytes to the egress counter of that project
-func (m *Module) AddEgress(project string, bytes uint64) {
-	m.lock.RLock()
-	defer m.lock.RUnlock()
-
-	// Return if the metrics module or bandwidth measurement is disabled
-	if !m.config.IsEnabled || m.config.DisableBandwidth {
+// AddEventingType counts the number of time a particular event type is called
+func (m *Module) AddEventingType(eventingType string) {
```
YourTechBud, line 11: Take the project here too

gateway/utils/metrics/operations.go
```diff
@@ -7,36 +7,37 @@ import (
...
+		m.eventing[eventingType] = 1
+		return
+	}
+	m.eventing[eventingType] = value + 1
 }
 
-// AddIngress add the bytes to the ingress counter of that project
-func (m *Module) AddIngress(project string, bytes uint64) {
+// AddFunctionOperation counts the number of time a particular function gets invoked
+func (m *Module) AddFunctionOperation(project string) {
```
YourTechBud, line 28: Take the service and endpoint name here as we

gateway/utils/metrics/metrics.go
```diff
@@ -1,69 +1,76 @@
...
+	m.ssl = ssl
 }
 
-type bwMetrics struct {
-	ingressBW uint64
-	egressBW  uint64
+type metrics struct {
+	crud      sync.Map // key -> dbType:col; value -> *metricOperations
+	fileStore sync.Map // key -> storeType value -> *metricOperations
+	function  uint64
```
YourTechBud, line 70: This will be a map as well. You'll also need to add a map for eventing module

gateway/utils/metrics/projects.go
```diff
@@ -0,0 +1,125 @@
...
+	if c == nil {
+		return "", nil, nil, true
+	}
+
+	// Get the cluster size
+	clusterSize, err := m.syncMan.GetClusterSize(context.Background())
+	if err != nil {
+		clusterSize = 1
+	}
+	find := m.syncMan.GetClusterID()
```
YourTechBud, line 32: Shouldn't find be a map[string]interface{}

gateway/utils/metrics/sink.go
```diff
@@ -2,44 +2,40 @@ package metrics
...
 	}
-
-	return sink, nil
 }
 
 func (m *Module) flushMetrics(docs []interface{}) {
+	if len(docs) == 0 {
+		return
+	}
+	logrus.Println("docs of flush", docs)
```
YourTechBud, line 31: Remove this comment

runner/actions.go
```diff
@@ -28,15 +28,17 @@ func actionRunner(c *cli.Context) error {
 	outsideCluster := c.Bool("outside-cluster")
 
 	isDev := c.Bool("dev")
+	isMetricDisabled := c.Bool("is-metric-disabled")
```
YourTechBud, line 31: Lets keep the flags name consistent. `disable-metrics`

runner/main.go
```diff
@@ -45,6 +45,11 @@ func main() {
 			Name:  "start",
 			Usage: "Starts a runner instance",
 			Flags: []cli.Flag{
+				cli.BoolFlag{
+					Name:   "is-metric-disabled",
+					EnvVar: "IS_METRIC_DISABLED",
+					Usage:  "enable or disable metrics collection",
```
YourTechBud, line 51: Keep same as gateway

gateway/main.go
```diff
@@ -142,9 +137,9 @@ var essentialFlags = []cli.Flag{
...
 	// Flags for the metrics module
 	cli.BoolFlag{
-		Name:   "enable-metrics",
-		Usage:  "Enable the metrics module",
-		EnvVar: "ENABLE_METRICS",
+		Name:   "disable-metrics",
+		Usage:  "Disable anonymous metric collection",
+		EnvVar: "DISABLE_METRICS",
 	},
 	cli.BoolFlag{
```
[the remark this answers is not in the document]
sharadregoti: yes, we require this metrics flags to disable metrics collection.
by default we collect both the metrics, (configuration & number of operations)

gateway/modules/crud/sql/update.go
```diff
@@ -77,7 +77,9 @@ func (s *SQL) update(ctx context.Context, project, col string, req *model.Update
 			doc := make(map[string]interface{})
 			dates := make(map[string]interface{})
 			for k, v := range req.Find {
-				doc[k] = v
+				for _, newValue := range v.(map[string]interface{}) {
+					doc[k] = newValue
+				}
```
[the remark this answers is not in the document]
sharadregoti: it was a bug,
while performing an upsert operation,
now when the row doesn' exists insert operation takes place, but in that id value is not replaced with the actual value
instead it's a map
in insert id should be 1, but it is id =  { "$eq" : "1"} 
so i range over it get the actual value
image

gateway/modules/eventing/handle_staged.go
```diff
@@ -190,5 +190,6 @@ func (m *Module) invokeWebhook(ctx context.Context, rule config.EventingRule, ev
 	}
 
 	_ = m.crud.InternalUpdate(ctxLocal, m.config.DBType, m.project, utils.TableEventingLogs, m.generateProcessedEventRequest(eventDoc.ID))
+	m.metricHook(eventDoc.Type)
```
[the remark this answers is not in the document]
sharadregoti: you wanted to know which eventing type is called how many times,
this metrics hook helps collecting that

gateway/modules/eventing/operations.go
```diff
@@ -54,10 +54,14 @@ func (m *Module) QueueEvent(ctx context.Context, project, token string, req *mod
 				// clear channel
 				return nil, ctx.Err()
 			case result := <-responseChan:
+				m.metricHook(req.Type)
 				return result, nil
 			}
 		}
 	}
+	if err == nil {
```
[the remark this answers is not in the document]
sharadregoti: my bad, error will always be nil if the code reaches that point

gateway/utils/metrics/projects.go
sharadregoti: not needed, update scMetrics handles it

gateway/main.go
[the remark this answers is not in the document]
YourTechBud: Im talking about the sink, conn and the other flags

gateway/modules/eventing/operations_test.go
```diff
@@ -152,6 +153,7 @@ func TestModule_SendEventResponse(t *testing.T) {
 }
 
 func TestModule_QueueEvent(t *testing.T) {
+	metricHook := func(project, eventingType string) {}
```
YourTechBud, line 156: You should mock calls to this function

gateway/modules/filestore/operations.go
```diff
@@ -34,6 +34,7 @@ func (m *Module) CreateDir(ctx context.Context, project, token string, req *mode
 	err = m.store.CreateDir(req)
 	if err == nil {
```
YourTechBud, line 35: I'd suggest you check if err != nil here. Also make it inline

gateway/modules/filestore/operations.go
```diff
@@ -65,6 +66,7 @@ func (m *Module) DeleteFile(ctx context.Context, project, token string, path str
 	err = m.store.DeleteFile(path)
 	if err == nil {
```
YourTechBud, line 67: I'd suggest you check if err != nil here. Also make it inline

gateway/modules/filestore/operations.go
```diff
@@ -120,6 +123,7 @@ func (m *Module) UploadFile(ctx context.Context, project, token string, req *mod
 	err = m.store.CreateFile(req, reader)
 	if err == nil {
```
YourTechBud, line 124: I'd suggest you check if err != nil here. Also make it inline

gateway/modules/filestore/store.go
```diff
@@ -17,15 +17,17 @@ import (
 // Module is responsible for managing the file storage module
 type Module struct {
 	sync.RWMutex
-	store    FileStore
-	enabled  bool
-	auth     model.AuthFilestoreInterface
-	eventing model.EventingModule
+	store       FileStore
+	storeType   string
```
YourTechBud, line 21: No need to store this. It can be fetched from the store object

gateway/modules/schema/update_test.go
```diff
@@ -159,19 +159,6 @@ func TestSchema_ValidateUpdateOperation(t *testing.T) {
...
-			args: args{
-				dbType: "mongo",
-				col:    "tweet",
-				updateDoc: map[string]interface{}{
-					"$inc": map[string]interface{}{
-						"age": 6.34,
-					},
-				},
-			},
-		},
```
YourTechBud, line 174: Let this test case be. It should not throw an error

gateway/utils/file.go
```diff
@@ -17,6 +17,7 @@ import (
 func Unzip(src string, dest string) error {
 
 	r, err := zip.OpenReader(src)
+
```
YourTechBud, line 20: Remove the space

gateway/utils/metrics/helpers.go
```diff
@@ -2,106 +2,97 @@ package metrics
...
-		"ts":         *t,
-		"count":      count,
-		"driver":     dbType,
-		"node_id":    "sc-" + m.nodeID,
+	if value.delete > 0 {
+		docs = append(docs, m.createDocument(arr[0], arr[1], arr[2], module, utils.Delete, value.delete, t))
 	}
+
+	if value.batch > 0 {
+		docs = append(docs, m.createDocument(arr[0], arr[1], arr[2], module, utils.Batch, value.batch, t))
```
YourTechBud, line 61: We shouldn't have a separate batch type. Instead let count the number of update, create and delete in each batch

gateway/utils/metrics/sink.go
```diff
@@ -2,44 +2,40 @@ package metrics
...
 )
 
+const metricsUpdaterInterval = 30 * time.Second
+
 func (m *Module) routineFlushMetricsToSink() {
-	ticker := time.NewTicker(1 * time.Second)
+	ticker := time.NewTicker(metricsUpdaterInterval)
 
 	for range ticker.C {
+		logrus.Println("executing ticker")
```
YourTechBud, line 17: remove this

gateway/utils/syncman/operations.go
```diff
@@ -229,12 +242,21 @@ func (s *Manager) GetProjectConfig(projectID string) ([]interface{}, error) {
 	defer s.lock.RUnlock()
 
 	// Iterate over all projects stored
+	v := []interface{}{}
 	for _, p := range s.projectConfig.Projects {
+		if projectID == "-" {
```
YourTechBud, line 247: use `*` instead of `~`

gateway/main.go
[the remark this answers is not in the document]
sharadregoti: nope

gateway/utils/metrics/projects.go
```diff
@@ -9,103 +9,94 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spaceuptech/space-api-go/types"
 
+	"github.com/spaceuptech/space-cloud/gateway/config"
 	"github.com/spaceuptech/space-cloud/gateway/utils"
 )
 
-func (m *Module) generateMetricsRequest() (string, map[string]interface{}, map[string]interface{}, bool) {
+func (m *Module) generateMetricsRequest(project *config.Project, ssl *config.SSL) (string, string, map[string]interface{}, map[string]interface{}) {
```
YourTechBud, line 16: Good to name all parameters
```suggestion
func (m *Module) generateMetricsRequest(project *config.Project, ssl *config.SSL) (clusterID, projectID string, set map[string]interface{}, min map[string]interface{}) {
```

PR #904, opened 2020-04-02
changed invocation log time to UTC

Jayesh333:


YourTechBud:
No need for this. We are adding the createdAt directive

Issue #899, opened 2020-03-31
Changes in security rules of eventing module does not get reflected immediatley 


Jayesh333:
### Describe the bug
After changing the security rule of eventing module, Space Cloud does not behave according to the new security rules. However, if I restart the space cloud, then it behaves according to the new rules. Not sure why is that.

### Expected behaviour
Space Cloud should immediately start behaving as per as the new rules after saving new rules.

### Actual behaviour 
Space Cloud still behaves as per old rules

### Steps to reproduce
Change the default eventing security rule from `allow` to `deny` and hit `Save`. Queue an event via Postman. It gets queued. (**Note:** It shouldn't have got queued because of `deny` rule). Restart space cloud and try queueing event once again from Postman. It gets denied this time 

### Your environment
- Space Cloud version: v0.17.0
- OS: Ubuntu

>If this bug restricts your use of space-cloud, give it a thumbs up reaction, so that we can determine which bugs need to be fixed immediately.
>👍

Jayesh333:
Done. Coming in `v0.17.0`

Issue #902, opened 2020-03-31
Queue event request gets stuck indefinitely


Jayesh333:
### Describe the bug
While queuing an event whose config is not specified, the request for queuing event gets stuck indefinitely. Also during this time when the queue request is stuck, if I try to change the config of any event trigger rule, then that also gets stuck. I feel this is related to locks.

### Expected behaviour
The request should pass or fail without getting stuck

### Actual behaviour 
The request gets stuck indefinitely.

### Steps to reproduce
Create a project from scratch and add a database. Don't create any event trigger rule. Queue an asynchronous event with any type and the request will get stuck.

### Your environment
- Space Cloud version: v0.17.0

>If this bug restricts your use of space-cloud, give it a thumbs up reaction, so that we can determine which bugs need to be fixed immediately.
>👍

Jayesh333:
Done. Coming in `v0.17.0`

Issue #901, opened 2020-03-31
Incorrect error logs in the eventing module 


Jayesh333:
### Describe the bug
Space Cloud logs the following error to the console while queueing a custom asynchronous event:
```bash
error sending synchronous event response to client unable to find channel in map for batch 1ZtaLIJEgqWDxLMuiswwx0venkl--auto-1ZtaKblxMnLbDLuVAceSxvMaYvR
```

### Expected behaviour
It shouldn't have tried sending the response back to the client and shouldn't have logged this error as the event queued was asynchronous

### Actual behaviour 
It logs the above specified error.

### Steps to reproduce
Queue an asynchronous event and check the console logs

### Your environment
- Space Cloud version: v0.17.0

>If this bug restricts your use of space-cloud, give it a thumbs up reaction, so that we can determine which bugs need to be fixed immediately.
>👍

Jayesh333:
Fixed. Coming in `v0.17.0`

PR #925, opened 2020-04-09
Ce get quotas endpoint

### LINKED ISSUE #916: Prevent creating multiple projects in open source binary
The open-source binary also has the `projects` config as an array which lets user create multiple projects. However only one of them is usable because of the way open-source binary works. But it might take user some time to realize this and all his efforts on the extra project he created might go waste. So the gateway should return error while creating a project if one project already exists

sharadregoti:
Added get quotas & credentials endpoint
this also fixes #916


YourTechBud:
Conflicts and tests are failing @sharadregoti

gateway/utils/server/server.go
```diff
@@ -32,16 +32,16 @@ type Server struct {
 }
 
 // New creates a new server instance
-func New(nodeID, clusterID, advertiseAddr, storeType, runnerAddr, artifactAddr string, removeProjectScope bool, metricsConfig *metrics.Config) (*Server, error) {
+func New(nodeID, clusterID, advertiseAddr, storeType, runnerAddr string, removeProjectScope bool, metricsConfig *metrics.Config, adminUserInfo *config.AdminUser, adminConfig *config.Admin) (*Server, error) {
```
YourTechBud, line 35: Dont pass admin config in new. Use the server.SetConfig instead.

YourTechBud:
Update the `space-api-go` version to `v0.17.0` in `go.mod` of space-cli, gateway and runner

YourTechBud:
Fixes #899 #901 #902. @Jayesh333 you'll need to confirm this

PR #941, opened 2020-04-16
metrics fix

sharadregoti:


gateway/utils/metrics/projects.go
```diff
@@ -97,7 +98,7 @@ func (m *Module) generateMetricsRequest(project *config.Project, ssl *config.SSL
 func (m *Module) updateSCMetrics(clusterID, projectID string, set, min map[string]interface{}) {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
-	result, err := m.sink.Upsert("config_metrics").Where(types.And(types.Cond("id", "==", clusterID), types.Cond("project", "==", projectID))).Set(set).Min(min).Apply(ctx)
+	result, err := m.sink.Upsert("config_metrics").Where(types.And(types.Cond("cluster_id", "==", clusterID), types.Cond("project", "==", projectID))).Set(set).Min(min).Apply(ctx)
```
YourTechBud, line 101: Why don't you just put the where clause on the id field?

gateway/main.go
```diff
@@ -54,7 +54,7 @@ var essentialFlags = []cli.Flag{
 		Name:   "cluster",
 		Usage:  "The cluster id to start space-cloud with",
 		EnvVar: "CLUSTER_ID",
-		Value:  "default-cluster",
+		Value:  ksuid.New().String(),
```
YourTechBud, line 57: Is this okay? Are you able to specify a clusterid via an env variable? See if you custom id actually gets applied or not?

gateway/utils/metrics/projects.go
sharadregoti: ok

gateway/main.go
sharadregoti: ok, i'll test with env variable

sharadregoti: yes, it works

YourTechBud: Awesome

gateway/utils/metrics/projects.go
```diff
@@ -34,7 +35,8 @@ func (m *Module) generateMetricsRequest(project *config.Project, ssl *config.SSL
 
 	modules := project.Modules
 	set["project"] = project.ID
-
+	id := fmt.Sprintf("%s--%s", clusterID, projectID)
+	set["id"] = id
```
YourTechBud, line 39: Remove this from set since its already in where

gateway/utils/metrics/projects.go
```diff
@@ -34,7 +35,8 @@ func (m *Module) generateMetricsRequest(project *config.Project, ssl *config.SSL
 
 	modules := project.Modules
 	set["project"] = project.ID
```
YourTechBud, line 37: Set the clusterid and project where you create the set object

PR #945, opened 2020-04-19
added json support for mysql

sharadregoti:


YourTechBud:
@sharadregoti did you try querying and updating the jsob column in mysql?
Try these to test things out.
- update json column
- confirm if json column returns properly as an object when ypu query.
- where clause on nested object field using _contains
- where clause on nested array field using _contains

sharadregoti:
i have tested all the cases you mentioned above, all of the above works

PR #964, opened 2020-05-06
added format error

prithvi081099:


gateway/modules/auth/match.go
```diff
@@ -77,7 +77,7 @@ func (m *Module) matchFunc(ctx context.Context, rule *config.Rule, MakeHTTPReque
 
 	scToken, err := m.GetSCAccessToken()
 	if err != nil {
-		return err
+		return formaterror(rule, err)
 	}
 
 	var result interface{}
```
YourTechBud, line 83: Pass matchHTTPRequest as the error in format error

gateway/modules/auth/match.go
```diff
@@ -94,11 +94,11 @@ func (m *Module) matchQuery(ctx context.Context, project string, rule *config.Ru
...
 	if err != nil {
-		return nil, err
+		return nil, formaterror(rule, err)
 	}
 	err = utils.StoreValue("args.result", data, args)
 	if err != nil {
-		return nil, err
+		return nil, formaterror(rule, err)
 	}
 	return m.matchRule(ctx, project, rule.Clause, args, nil)
```
YourTechBud, line 103: Pass the error returned by match rule inside format error. Return the postAction object as is

gateway/modules/auth/match.go
```diff
@@ -128,7 +128,7 @@ func (m *Module) matchOr(ctx context.Context, projectID string, rule *config.Rul
 		}
 	}
 	// if condition is not satisfied -> return empty model.PostProcess and error
-	return nil, ErrIncorrectMatch
+	return nil, formaterror(rule, ErrIncorrectMatch)
 }
 
 func match(rule *config.Rule, args map[string]interface{}) error {
```
YourTechBud, line 134: Dont return match errors from matchString, Match Number,etc as is. Pass it through format error

gateway/modules/auth/match_types.go
```diff
@@ -13,32 +13,32 @@ func matchString(rule *config.Rule, args map[string]interface{}) error {
 	var f2 string
 	f1String, ok := rule.F1.(string)
 	if !ok {
-		return ErrIncorrectRuleFieldType
+		return formaterror(rule, ErrIncorrectRuleFieldType)
```
YourTechBud, line 16: Remove all format error statements from this file. It will get handled in the `match` rule handler itself

gateway/modules/auth/errors.go
```diff
@@ -10,3 +15,16 @@ var ErrIncorrectRuleFieldType = errors.New("auth: Incorrect rule field type")
 
 // ErrIncorrectMatch is thrown when the field type of a rule is of incorrect type
 var ErrIncorrectMatch = errors.New("auth: The two fields do not match")
+
+//Formaterror check whether error is provided in config.Rule
+func formatError(rule *config.Rule, err error) error {
+	if rule.Name == "" {
```
YourTechBud, line 21: The first thing you should do is check if err you receive as a parameter is nil. Simply return nil if thats the case

gateway/modules/auth/match.go
```diff
@@ -67,7 +67,8 @@ func (m *Module) matchRule(ctx context.Context, project string, rule *config.Rul
 		return matchHash(rule, args)
 
 	default:
-		return nil, ErrIncorrectMatch
+
+		return nil, formatError(rule, fmt.Errorf("invalid rule type (%s) provided", ErrIncorrectMatch))
```
YourTechBud, line 71:
```suggestion
		return nil, formatError(rule, fmt.Errorf("invalid rule type (%s) provided", rule.Rule))
```

gateway/modules/auth/errors.go
YourTechBud, line 21: We should first check if the err is nil. Simply return nil if it is

gateway/modules/auth/match.go
```diff
@@ -128,22 +131,22 @@ func (m *Module) matchOr(ctx context.Context, projectID string, rule *config.Rul
...
-		return matchNumber(rule, args)
+		return formatError(rule, matchNumber(rule, args))
 
 	case "bool":
-		return matchBool(rule, args)
+		return formatError(rule, matchBool(rule, args))
 	}
 
-	return ErrIncorrectMatch
+	return formatError(rule, ErrIncorrectMatch)
```
YourTechBud, line 149:
```suggestion
	return formatError(rule, fmt.Errorf("invalid variable data type (%s) provided", rule.Type))
```

Issue #497, opened 2019-11-10
Remove `remove-project-scope` flag


HeyItsJs:
Right now we have implemented a quick fix for one of the weird errors MySQL was giving when database name was provided in the query. 

We should solve this problem in such a way that there should be no need of providing this flag. The table name to be used in query should be generated based on whether the user has provided the db name in the connection string rather than the value of this flag.

YourTechBud:
@Jayesh333 i think this is an important feature to have. It's useful when the user wants to use a db name which is not the project id. Let's keep it?

Jayesh333:
The problem with `remove-project-scope` flag as of now is that it is defined globally. Whereas, it should be defined at a database level.

And more importantly Space Cloud should be able to detect whether to remove the project scope or not without the user telling him. If the connection string contains db name then it should remove the project scope otherwise use project scope.

Jayesh333:
After thinking for a while, I have this solution in mind:

The current implementation for `getTableName` function (and other similar functions), which is used to calculate the name of the table in the database queries is:

```go
func getTableName(project, table string, removeProjectScope bool) string {
	if removeProjectScope {
		return table
	}

	return project + "." + table
}
```

We can change its implementation to this:
```go
func getTableName(table, scope string) string {
	if scope != nil {
		table
	}

	return scope + "." + table
}
```

This `scope` will be set at the `database` level while initializing/setting the config of that database. 

**For MySQL:** While initializing we can check whether the `conn` string contains the database name or not to set the `scope` accordingly. If the database name is present in the connection string, we will set the `scope` to `nil`, so that the table name is returned in the helper functions. If the database name is not present, we will set the `scope` to `project-id`.

**New possibilities unlocked because of `scope`:**

This `scope` can prove to be a very powerful tool. For example, we can use it to configure Space Cloud to use `public` or some other `schema` in case of `Postgres`. (Right now, there is no easy way to do it). Having it as a config parameter for the database, we can now also ask the user upfront, which database or schema is he interested to use. And also allow them to change this config on the fly without requiring to re run space cloud.

YourTechBud:
Done

PR #971, opened 2020-05-09
Ce support for specifying database name

sharadregoti:


gateway/config/config.go
```diff
@@ -69,6 +69,7 @@ type Crud map[string]*CrudStub // The key here is the alias for database type
 // CrudStub holds the config at the database level
 type CrudStub struct {
 	Type            string                    `json:"type,omitempty" yaml:"type"` // database type
+	Name            string                    `json:"name,omitempty" yaml:"name"` // name of the logical database or schema name according to the database type
```
YourTechBud, line 72: We should call this field `database` or `dbName`

gateway/modules/crud/crud.go
```diff
@@ -81,14 +80,23 @@ func (m *Module) SetHooks(hooks *model.CrudHooks, metricHook model.MetricCrudHoo
 	m.metricHook = metricHook
 }
 
-func (m *Module) initBlock(dbType utils.DBType, enabled bool, connection string) (Crud, error) {
+func (m *Module) initBlock(dbType utils.DBType, enabled bool, connection, name string) (Crud, error) {
```
YourTechBud, line 83: Don't call it name, call it `dbName`

gateway/modules/crud/crud.go
```diff
@@ -81,14 +80,23 @@ func (m *Module) SetHooks(hooks *model.CrudHooks, metricHook model.MetricCrudHoo
...
 	case utils.Mongo:
-		return mgo.Init(enabled, connection)
+		return mgo.Init(enabled, connection, name)
 	case utils.EmbeddedDB:
 		return bolt.Init(enabled, connection)
 	case utils.MySQL, utils.Postgres, utils.SQLServer:
-		return sql.Init(dbType, enabled, m.removeProjectScope, connection)
+		c, err := sql.Init(dbType, enabled, connection, name)
+		if err == nil && enabled {
+			if err := c.CreateDatabaseIfNotExist(context.Background(), name); err != nil {
```
YourTechBud, line 92: We shouldn't use context.Background here.

gateway/modules/crud/crud.go
```diff
@@ -57,7 +56,7 @@ type Crud interface {
 	RawQuery(ctx context.Context, query string, args []interface{}) (int64, interface{}, error)
 	GetCollections(ctx context.Context, project string) ([]utils.DatabaseCollections, error)
 	DeleteCollection(ctx context.Context, project, col string) error
```
YourTechBud, line 58: We can remove the project parameter from all methods since nobody is using it

YourTechBud: We can also remove it from all the exposed method in `operations.go` and `internal.go`

gateway/config/config.go
sharadregoti: i 'll rename it to dbName,
i had given a generic name as it can be db name or schema name.

YourTechBud:
@sharadregoti everything tested?

sharadregoti:
yes

YourTechBud:
Fixes #497

PR #987, opened 2020-05-14
Filestore secret

khannaabhi:


gateway/modules/filestore/store.go
```diff
@@ -65,6 +69,14 @@ func (m *Module) SetConfig(conf *config.FileStore) error {
 		}
 	}
 
+	// set fileStore secret for aws and gcp
+	connection := strings.Split(conf.Conn, ".")
```
sharadregoti, line 73: 1) You are using the wrong variable for splitting, you should use **conf.Secret**
2) Make a function that does the splitting, it will return secretName,secretKey,isSecretExists
3) As the function is same for both crud & filestore. Write the above function in utils folder

gateway/modules/filestore/store.go
```diff
@@ -65,6 +69,14 @@ func (m *Module) SetConfig(conf *config.FileStore) error {
...
+	if connection[0] == "secrets" {
+		if err := setFileSecret(utils.FileStoreType(conf.StoreType), connection[2], conf.Secret); err != nil {
```
sharadregoti, line 75: Before the setting connection
You shoudl fetch secrets from runner.

PR #990, opened 2020-05-14
Get project

### LINKED ISSUE #989: All space-cli getters throws weird errors if project flag isn't provided
Running `space-cli get <>` without the `--project` flag throws unpredictable errors. We should have a function called `getProjectID()` which returns the projectID and a boolean (specifying if the projectID) exists are not. This function will attempt to get the projectId from viper.
All getters can call this function and check if the projectId has been provided

prithvi081099:
Fixes #989


space-cli/cmd/modules/deploy/commands.go
```diff
@@ -42,7 +42,11 @@ func Commands() []*cobra.Command {
 }
 
 func actionDeploy(cmd *cobra.Command, args []string) error {
-	projectID := viper.GetString("project")
+	projectID, check := utils.GetProjectID()
+	if !check {
+		_ = utils.LogError("Project not specified in flag", nil)
+		return nil
+	}
```
YourTechBud, line 49: Revert back to previous thing. This isn't required

PR #1029, opened 2020-05-26
cli command return error

### LINKED ISSUE #1027: Space CLI commands should return an error
Currently we simply print the errors but don't return them. This causes false positives in the CI/CD pipeline. For eg. `space-cli apply` failed but it still returns as successful.
### How can we solve it?
Simply return the errors
>👍

prithvi081099:
Fixes #1027


YourTechBud:
To be clear. I asked to add the `SilenceUsage` to the root command only and `SilentErrors` from top level subcommands

PR #1050, opened 2020-06-06
Mysql bool return value

### LINKED ISSUE #758: MySQL returns 1/0 instead of true/false for Boolean values
Mysql stores boolean value in the form of a single bit 1/0. There's no boolean type in MySQL.

However, we provide a unified API across all database and thus the user should be able to receive true/false for Boolean type fields while querying as it does in other databases. As of now, Space Cloud returns 1/0 to the user in case of MySQL for boolean fields. It should convert it into Boolean value before returning it to the user.

KavishShah09:
Fixes #758


sharadregoti:
@KavishShah09 add test cases

PR #1058, opened 2020-06-09
added collation to sql database

### LINKED ISSUE #1031: Mysql & Sql-Server does a case insensitive search on column with type string
### Describe the bug
Assume table customers with schema
**type customers {
id : ID! @primary
name : String!
}**
Having 2 rows
**id                    name
1                     john
2                     John**

Executing the SQL query **Select * from customers where name = 'john'** through MySQL shell returns 2 rows, as by default MySQL does a case insensitive search on a column with type text or varchar.
Same applies for SQL-server.

But if we execute the following graphql query through space-cloud
**query {
customers(where : {name : john}) @mysql {
id 
name
}**
Returns only 1 row instead of 2, as space cloud does some post-processing on the result( in utils.Validate() function )

### Expected behaviour
Even though MySQL & SQL-server do case insensitive search. space cloud should force case sensitive search by adding **collate** field while creating tables in space cloud.

For case insensitive search, developers can use **regex**

### Your environment
- Space Cloud version:
- OS:
- The client API version (if applicable):
- Browser (if applicable):

>👍

prithvi081099:
Fixes #1031


sharadregoti:
@prithvi081099 your tests are failing

sharadregoti:
also add collattion at dabase level

gateway/modules/schema/helpers.go
```diff
@@ -93,6 +93,9 @@ func (c *creationModule) addNotNull() string {
 	case utils.Postgres:
 		return "ALTER TABLE " + c.schemaModule.getTableName(dbType, c.logicalDBName, c.TableName) + " ALTER COLUMN " + c.ColumnName + " SET NOT NULL"
 	case utils.SQLServer:
+		if c.columnType == "varchar("+model.SQLTypeIDSize+")" || c.columnType == "varchar(max)" || c.columnType == "text" {
+			return "ALTER TABLE " + c.schemaModule.getTableName(dbType, c.logicalDBName, c.TableName) + " ALTER COLUMN " + c.ColumnName + " " + c.columnType + " collate Latin1_General_CS_AS NOT NULL"
+		}
```
sharadregoti, line 98: remove text type in if case, Do this for all

gateway/utils/constants.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 // BuildVersion is the current version of Space Cloud
-const BuildVersion = "0.19.0"
+const BuildVersion = "0.18.0"
```
YourTechBud, line 8:
```suggestion
const BuildVersion = "0.18.1"
```

gateway/modules/schema/helpers.go
```diff
@@ -93,6 +93,9 @@ func (c *creationModule) addNotNull() string {
 	case utils.Postgres:
 		return "ALTER TABLE " + c.schemaModule.getTableName(dbType, c.logicalDBName, c.TableName) + " ALTER COLUMN " + c.ColumnName + " SET NOT NULL"
 	case utils.SQLServer:
+		if c.columnType == "varchar("+model.SQLTypeIDSize+")" || c.columnType == "varchar(max)" {
+			return "ALTER TABLE " + c.schemaModule.getTableName(dbType, c.logicalDBName, c.TableName) + " ALTER COLUMN " + c.ColumnName + " " + c.columnType + " collate Latin1_General_CS_AS NOT NULL"
+		}
```
YourTechBud, line 98: Whats the point of this? Maybe you should just check if there is a `varchar` prefix?

sharadregoti: yep

YourTechBud: Okay. So @prithvi081099 you'll need to change this everywhere

prithvi081099: I have made the changes @YourTechBud and @sharadregoti

gateway/utils/constants.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 // BuildVersion is the current version of Space Cloud
-const BuildVersion = "0.18.0"
+const BuildVersion = "0.18.1"
```
sharadregoti, line 8:
```suggestion
const BuildVersion = "0.18.0"
```

Issue #1079, opened 2020-06-14
GraphQL API for services Not working after upgrading from v17 to v18


sharadregoti:
### Describe the bug
Graphql API stops working after upgrading to v18 from v17. I get this error message when i try to query an endpoint
 **"message": "Invalid endpoint kind () provided"**
This is because a new field has been added to the function module struct called kind. Which gets empty value after updating to v18

### Your environment
- Space Cloud version:
- OS:
- The client api version (if applicable):
- Browser (if applicable):

>If this bug restricts your use of space-cloud, give it a thumbs up reaction, so that we can determine which bugs need to be fixed immediately.
>👍

HeyItsJs:
Delete the `$HOME/space-cloud` folder and try again. Its because you have old mission control through which you configured space cloud. The same had happened with me.

Don't forget to close this issue if it gets solved @sharadregoti

PR #1101, opened 2020-06-20
Solves Remote Service Bug After Upgrading v17 to v18

sharadregoti:


YourTechBud:
@sharadregoti fix the tests

sharadregoti:
fixes #1079

PR #1105, opened 2020-06-23
Aggregations on linked tables

### LINKED ISSUE #1061: Aggregations on linked tables
Right now aggregations only work on base tables and not on linked tables. 

For example, let's say we have these two tables:
1] users:
```graphql
type users {
  id: ID! @primary
  name: String!
  expenditures: [expenditures] @link(table: expenditures, from: id, to: user_id)
}
```
2] expenditures
```graphql
type expenditures {
  id: ID! @primary
  category: ID!
  amount: Integer!
  user_id: ID! @foreign(table: users, field: id)
  user: users @link(table: users, from: user_id, to: id)
}
```
If I want to fetch a user's aggregated expenditures, I can do it like this:
```graphql
{
  expenditures(where: {user_id: "1"}) @db {
    aggregate {
      count {
        id
      }
      max {
        amount
      }
      min {
        amount
      }
      avg {
        amount
      }
      sum {
        amount
      }
    }
  }
}

```

However, I can't perform aggregations on expenditures table when it is used as a linked table in the query like these:

```graphql
{
  users @db {
    id
    name
    expenditures {
      aggregate {
        count {
          id
        }
        max {
          amount
        }
        min {
          amount
        }
        avg {
          amount
        }
        sum {
          amount
        }
      }
    }
  }
}
```

If I execute the above query, I get `aggregate` field as null.

It will be great if we can perform aggregations on linked tables as well.

prithvi081099:
Fixes #1061


gateway/utils/graphql/read.go
```diff
@@ -20,12 +20,29 @@ func (graph *Module) execLinkedReadRequest(ctx context.Context, field *ast.Field
 		return
 	}
 
+	req.Aggregate, err = extractAggregate(field)
+	if err != nil {
+		cb("", "", nil, err)
+		return
+	}
+	var hasOptions bool
```
sharadregoti, line 28: also extract the group by clause, check generateReadRequest function( )

gateway/utils/graphql/read.go
```diff
@@ -20,12 +20,29 @@ func (graph *Module) execLinkedReadRequest(ctx context.Context, field *ast.Field
...
+	req.Options, hasOptions, err = generateOptions(field.Arguments, store)
+	if err != nil {
+		cb("", "", nil, err)
+		return
+	}
+	// if distinct option has been set then set operation to distinct from all
+	if hasOptions && req.Options.Distinct != nil {
+		req.Operation = utils.Distinct
+	}
```
sharadregoti, line 37: @YourTechBud should we support options for linked requests ?

sharadregoti: @YourTechBud ?

YourTechBud: No. We shouldn't

PR #1107, opened 2020-06-24
Count Function works without selection set for aggregations (SQL)

### LINKED ISSUE #1063: count function in aggregations should work even without a selection set
For using aggregation functions (`min`, `max`, `sum`, etc.) we need to provide a selection set inside the aggregation functions denoting the fields on which we want to perform the aggregation function.

For example:

```graphql
{
  expenditures @db {
    aggregate {
      sum {
        amount
      }
    }
  }
}
```

For example, the above aggregation query means that we want the sum of the `amount` inside the `expenditures` table. 

If we don't provide the field inside the aggregation function (`amount` in the above example), then we get the `null` as the value for `aggregate`. This makes complete sense as the `sum` function needs a column/field.

However, for the `count` function there should be no such restriction, we should allow a `count` aggregation even if a field is not specified inside it.  Meaning this query should ideally work:

```graphql
{
  expenditures @db {
    aggregate {
      count
    }
  }
}
```

In the SQL world, this is equivalent of:
  
```sql
select count(*) from <table-name>;
```

Basically, we should use `*` if a field is not specified inside the `count` function

KavishShah09:
Fixes #1063 
Task linked : https://app.clickup.com/t/3bttq9


gateway/utils/graphql/read.go
```diff
@@ -182,6 +182,12 @@ func extractAggregate(v *ast.Field) (map[string][]string, error) {
 			if ok {
 				return nil, utils.LogError(fmt.Sprintf("Cannot repeat the same function (%s) twice. Specify all columns within single function field", functionField.Name.Value), "graphql", "extractAggregate", nil)
 			}
+
+			if functionField.Name.Value == "count" && functionField.SelectionSet == nil {
+				functionMap[functionField.Name.Value] = []string{""}
+				return functionMap, nil
```
sharadregoti, line 188: I think you should use **continue** here, there can other functions also.
Also test it with multiple functions
aggregate {
   count
   sum{
     col_a
     col_b
}
}

gateway/modules/crud/sql/read.go
```diff
@@ -248,11 +248,15 @@ func (s *SQL) readexec(ctx context.Context, sqlString string, args []interface{}
 						funcValue, ok := funcMap[functionName]
 						if !ok {
 							// set new function
-							funcMap[functionName] = map[string]interface{}{columnName: value}
+							if columnName == "" {
```
sharadregoti, line 251: Write a comment
// NOTE: This case occurs for count function with no column name (using * operator instead)

KavishShah09:
@sharadregoti this is tested for multiple function and multiple columns in them.

PR #1113, opened 2020-06-28
Link across database

### LINKED ISSUE #1018: Add support for links across databases.
Currently we need to use on the fly joins and manual mutations for working with linked tables across multiple databases

### How can we solve it?
There should be a `db` argument to the `@link` directive which sets the dbAlias field in the `LinkTable Properties`. This alone will take care of linked queries.
We also have to update the mutation  mechanism.

>👍

prithvi081099:
Fixes #1018


sharadregoti:
@YourTechBud how can we test this pr ?
As open source version doesnt allow to have multiple databases ?

YourTechBud:
@sharadregoti lets just merge it for now. We'll test it later

YourTechBud:
@prithvi081099 run `go mod tidy` and push it again

sharadregoti:
@YourTechBud does this code work for both query & mutations ?

YourTechBud:
Yes

gateway/utils/graphql/mutation.go
```diff
@@ -116,7 +118,9 @@ func (graph *Module) handleMutation(ctx context.Context, node ast.Node, token st
 
 		// Add the request to the number of requests available for that database
 		r = append(r, generatedRequests...)
-		reqs[dbAlias] = r
+		for _, v := range r {
```
YourTechBud, line 121: Range over `generatedRequests`. We don't need `r`

gateway/utils/graphql/mutation.go
```diff
@@ -32,6 +32,7 @@ func (graph *Module) generateAllReq(ctx context.Context, field *ast.Field, token
 			}
 			result.Type = string(utils.Delete)
 			result.Col = col
+			result.DBAlias, _ = graph.GetDBAlias(field)
```
YourTechBud, line 35: Instead of doing this, why dont you just pass the dbAlias as a parameter in this function

PR #1139, opened 2020-07-06
Add better usage hints, alias and example for generate sub commands

### LINKED ISSUE #1127: Provide better hints when using space-cli generate incorrectly
I used the following space-cli command to generate remote-services:

```bash
space-cli generate remote-services --project myproject
```

It gives me the following error:
```
ERRO[0000] incorrect number of arguments                
Error: incorrect number of arguments
```

As a next step that any developer would take, I ran the following command:
```bash
space-cli generate remote-services --help
```
And here's what I get 😐:
```bash
Usage:
  space-cli generate remote-services [flags]

Flags:
  -h, --help   help for remote-services

Global Flags:
      --log-level string   Sets the log level of the command (default "info")
      --project string     The project id to perform the options in
```   

This isn't really helpful. Ideally, the error message should have only said what argument it required in the first place. I this was some other tool, I would have dropped off at this point as I faced a roadblock which I couldn't solve.

prakashmishra1598:
Fixes #1127. Provides better usage instructions and and Example while executing the command


prakashmishra1598:
@YourTechBud, can you review?

YourTechBud:
@prakashmishra1598 can we do this for all the generators we have. You can find a list of all `generators` here: https://github.com/spaceuptech/space-cloud/blob/v0.19.0/space-cli/cmd/modules/generator.go

space-cli/cmd/modules/remote-services/commands.go
```diff
@@ -10,8 +10,9 @@ import (
 func GenerateSubCommands() []*cobra.Command {
 
 	var generateService = &cobra.Command{
-		Use:  "remote-services",
-		RunE: actionGenerateService,
+		Use:     "remote-services [config-file-name]",
```
YourTechBud, line 13:
```suggestion
		Use:     "remote-service [path to config file]",
```

prakashmishra1598:
@YourTechBud Ack. I'll let you know when the changes are done

space-cli/cmd/modules/database/commands.go
```diff
@@ -10,23 +10,30 @@ import (
...
+		Use:  "db-schema [path to config file]",
 		RunE: actionGenerateDBSchema,
+		Aliases: []string{"db-schemas"},
+		Example: "space-cli generate db-schema config.yaml --project myproject --log-level info",
 	}
 
 	var generatePreparedQuery = &cobra.Command{
 		Use:  "db-prepared-query",
 		RunE: actionGenerateDBPreparedQuery,
+		Example: "space-cli generate db-prepared-query config.yaml --project myproject --log-level info",
```
YourTechBud, line 36: You missed the Alias here

space-cli/cmd/modules/project/commands.go
```diff
@@ -11,8 +11,10 @@ import (
 func GenerateSubCommands() []*cobra.Command {
 
 	var generateService = &cobra.Command{
```
YourTechBud, line 13: Lets rename this to generateProject

YourTechBud:
@sharadregoti can you confirm if the usage for `space-cli generate service` command is accurate?

space-cli/cmd/modules/database/commands.go
prakashmishra1598: I wasn't sure if I should add an alias here. The cli doesn't ask the user if he wants to add more prepared queries after adding one, so I thought that adding an alias for it would be unecessary

YourTechBud: Thats a valid point. But i think all generators behave that way. I proposed adding Alias to makenit consistent with the getter APIs

space-cli/cmd/modules/services/commands.go
```diff
@@ -132,7 +134,7 @@ func actionGetServices(cmd *cobra.Command, args []string) error {
 func actionGenerateService(cmd *cobra.Command, args []string) error {
 	// get filename from args in which service config will be stored
 	if len(os.Args) != 4 {
```
sharadregoti, line 136:
```suggestion
	if len(args) != 1 {
```

space-cli/cmd/modules/services/commands.go
```diff
@@ -132,7 +134,7 @@ func actionGetServices(cmd *cobra.Command, args []string) error {
...
-		return utils.LogError("incorrect number of arguments", nil)
+		return utils.LogError("incorrect number of arguments. Use -h to check usage instructions", nil)
 	}
 	serviceConfigFile := os.Args[3]
```
sharadregoti, line 139:
```suggestion
	serviceConfigFile := args[0]
```

sharadregoti:
Yes, the usage is correct

space-cli/cmd/modules/remote-services/commands.go
```diff
@@ -10,8 +10,10 @@ import (
 func GenerateSubCommands() []*cobra.Command {
 
 	var generateService = &cobra.Command{
-		Use:  "remote-services",
-		RunE: actionGenerateService,
+		Use:     "remote-service [path to config file]",
```
prakashmishra1598, line 13: @sharadregoti @YourTechBud Can you check this part again and confirm the usage? I commit changes proposed by @sharadregoti

YourTechBud: Seems to be correct. @sharadregoti?

sharadregoti: yes

PR #1150, opened 2020-07-09
Add support to specify response headers in ingress routing

### LINKED ISSUE #1146: Add support to insert response headers in IngressRouting
Currently, we can set request headers only. In many cases (like to enable cors) we need to set certain response headers.

It would be great if we can do that directly from SpaceCloud

YourTechBud:
fixes #1146, #1151, #1152


YourTechBud:
@HeyItsJs this is done from my side. Test it and merge it.

Issue #1022, opened 2020-05-23
List files (S3) gives error


HeyItsJs:
If I use S3 as a file store and try to list the files, Space Cloud logs the following error on the console:

```bash
2020/05/22 17:56:02 http: panic serving 172.26.0.1:47434: runtime error: slice bounds out of range [1:0]
goroutine 1377 [running]:
net/http.(*conn).serve.func1(0xc0000f9ae0)
	/usr/local/go/src/net/http/server.go:1767 +0x139
panic(0x1fc3760, 0xc0009c38c0)
	/usr/local/go/src/runtime/panic.go:679 +0x1b2
github.com/spaceuptech/space-cloud/gateway/modules/filestore/amazons3.(*AmazonS3).ListDir(0xc000984300, 0xc0009487c0, 0xc000a771a0, 0xc000053ecc, 0x9, 0x2115948, 0x0)
	/build/modules/filestore/amazons3/read.go:33 +0x7e5
github.com/spaceuptech/space-cloud/gateway/modules/filestore.(*Module).ListFiles(0xc000838960, 0x265bea0, 0xc000a771a0, 0xc000053ecc, 0x9, 0x2115948, 0x0, 0xc0009487c0, 0x0, 0x0, ...)
	/build/modules/filestore/operations.go:91 +0x1a5
github.com/spaceuptech/space-cloud/gateway/utils/handlers.HandleRead.func1(0x2652620, 0xc00070efc0, 0xc0005d4c00)
	/build/utils/handlers/filestore.go:128 +0x781
net/http.HandlerFunc.ServeHTTP(0xc00080c450, 0x2652620, 0xc00070efc0, 0xc0005d4c00)
	/usr/local/go/src/net/http/server.go:2007 +0x44
github.com/gorilla/mux.(*Router).ServeHTTP(0xc0001f6000, 0x2652620, 0xc00070efc0, 0xc00095c700)
	/build/vendor/github.com/gorilla/mux/mux.go:212 +0xe2
github.com/spaceuptech/space-cloud/gateway/utils/server.(*Server).restrictDomainMiddleware.func1(0x2652620, 0xc00070efc0, 0xc00095c700)
	/build/utils/server/http.go:30 +0x18e
net/http.HandlerFunc.ServeHTTP(0xc0007ec9c0, 0x2652620, 0xc00070efc0, 0xc00095c700)
	/usr/local/go/src/net/http/server.go:2007 +0x44
[...]

Issue #1133, opened 2020-07-04
Returning docs in insert query doesn't return the docs that it has mutated internally


sharadregoti:
### Describe the bug
The returning docs in insert query doesn't return the docs that it has mutated internally
like adding default value if given in query, adding createdat or updateAt
### Your environment
- Space Cloud version:
- OS:
- Kubernetes or Docker:
- The client api version (if applicable):
- Browser (if applicable):

>If this bug restricts your use of space-cloud, give it a thumbs up reaction, so that we can determine which bugs need to be fixed immediately.
>👍

sharadregoti:
Closing this issue as it is a false positive

PR #1180, opened 2020-07-29
Minor bug fixes

### LINKED ISSUE #1100: Check if secrets is the same before initialising the crud module
Currently we only use the connection string and dbname in the crud config to check if the crud block needs to be reinitialised. This will cause a problem when using secrets since the secret value won't be available to the `IsSame` function.

What we need to do is update the conn string value from the secret and pass that to isSame. We also need to store the certificates inside each crud block and check if that is same as well.

**This issue can be solved after merging `0.18.1` in `0.19.0`.**

sharadregoti:
This fixes #1100 #1133 & some minor fixes

### DIFF
--- a/gateway/managers/syncman/syncman_crud.go
+++ b/gateway/managers/syncman/syncman_crud.go
@@ -148,7 +148,7 @@ func (s *Manager) GetPreparedQuery(ctx context.Context, project, dbAlias, id str
 			return http.StatusOK, []interface{}{&preparedQueryResponse{ID: id, DBAlias: dbAlias, SQL: preparedQuery.SQL, Arguments: preparedQuery.Arguments, Rule: preparedQuery.Rule}}, nil
 		}
 		preparedQuery := databaseConfig.PreparedQueries
-		var coll = make([]interface{}, 0)
+		coll := make([]interface{}, 0)
 		for key, value := range preparedQuery {
 			coll = append(coll, &preparedQueryResponse{ID: key, DBAlias: dbAlias, SQL: value.SQL, Arguments: value.Arguments, Rule: value.Rule})
 		}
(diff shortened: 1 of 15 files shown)


gateway/modules/crud/crud.go
```diff
@@ -148,10 +148,20 @@ func (m *Module) SetConfig(project string, crud config.Crud) error {
 			m.queries[getPreparedQueryKey(strings.TrimPrefix(k, "sql-"), id)] = query
 		}
 
+		// check if connection string starts with secrets
+		secretName, secretKey, isSecretExists := splitConnectionString(v.Conn)
+		if isSecretExists {
+			var err error
+			v.Conn, err = m.getSecrets(project, secretName, secretKey)
```
YourTechBud, line 155: You can't set this here. This will update the config object

gateway/modules/filestore/amazons3/read.go
```diff
@@ -23,14 +25,21 @@ func (a *AmazonS3) ListDir(req *model.ListFilesRequest) ([]*model.ListFilesRespo
...
 	})
+	if err != nil {
+		return nil, err
+	}
 
-	result := []*model.ListFilesResponse{}
+	if len(resp.Contents) == 0 {
+		utils.LogDebug("AWS list response is empty", "amazons3", "list-dir", nil)
+		return nil, nil
+	}
```
YourTechBud, line 40: Return an empty array

gateway/modules/crud/crud.go
```diff
@@ -148,10 +148,21 @@ func (m *Module) SetConfig(project string, crud config.Crud) error {
 			m.queries[getPreparedQueryKey(strings.TrimPrefix(k, "sql-"), id)] = query
 		}
 
+		// check if connection string starts with secrets
+		secretName, isSecretExists := splitConnectionString(v.Conn)
+		var connectionString = v.Conn
```
YourTechBud, line 153:
```suggestion
		connectionString := v.Conn
```

gateway/managers/syncman/syncman_crud.go
```diff
@@ -148,17 +148,17 @@ func (s *Manager) GetPreparedQuery(ctx context.Context, project, dbAlias, id str
 			return http.StatusOK, []interface{}{&preparedQueryResponse{ID: id, DBAlias: dbAlias, SQL: preparedQuery.SQL, Arguments: preparedQuery.Arguments, Rule: preparedQuery.Rule}}, nil
 		}
 		preparedQuery := databaseConfig.PreparedQueries
-		var coll []interface{} = make([]interface{}, 0)
+		var coll = make([]interface{}, 0)
```
YourTechBud, line 151:
```suggestion
		coll := make([]interface{}, 0)
```

YourTechBud:
Fixes #1022

PR #1185, opened 2020-08-01
GCP File store fixes

sharadregoti:


gateway/modules/filestore/gcpstorage/create.go
```diff
@@ -3,14 +3,20 @@ package gcpstorage
...
 )
 
 // CreateFile creates a file in GCPStorage
 func (g *GCPStorage) CreateFile(req *model.CreateFileRequest, file io.Reader) error {
-	wc := g.client.Bucket(g.bucket).Object(utils.JoinLeading(req.Path, req.Name, "/")).NewWriter(context.TODO())
+	req.Path = strings.TrimPrefix(req.Path, "/")
+	path := req.Path + "/" + req.Name
+	if len(req.Path) == 0 {
+		path = req.Name
+	}
```
YourTechBud, line 18: Logic is incorrect

gateway/modules/filestore/gcpstorage/delete.go
```diff
@@ -10,17 +10,17 @@ import (
 
 // DeleteFile deletes a file from GCPStorage
 func (g *GCPStorage) DeleteFile(path string) error {
-	// start of path must have single backslash
-	// if end is a backslash, it will delete a folder
-	if strings.HasSuffix(path, "/") {
-		return g.DeleteDir(path)
-	}
+	// trim / at the start and at the end
+	path = strings.TrimPrefix(strings.TrimSuffix(path, "/"), "/")
```
YourTechBud, line 14: Make sure we trim suffix too

sharadregoti: we are doing it. Look at the first param of trim prefix

gateway/modules/filestore/gcpstorage/create.go
sharadregoti: Look carefully I am checking the length of req.path after trim prefix.
if user only provides / . then because of trim prefix req.path will be empty in that case we have to just create a file inside the root directory

PR #1199, opened 2020-08-05
Implemented running gateway container as non root user

### LINKED ISSUE #1198: Run gateway image as non-root user
Currently all files needed to run gateway image are owned by root user. 
Also the mission-control package is unpacked into /root directory.

This setup makes it impossible to run gateway image on k8s clusters with tight security restrictions like:
- containers cannot be run as root ( non root user cannot create config file in /app dir and access mission-control files)
- container don't have access to internet (container cannot download mission-control after failing to read one in root dir)

### How can we solve it?
To overcome this, dedicated user should be created in the dockerfile and set as default user in the container. This user should own all the files and directories needed to run the application.

Happy to issue a PR if this gets and OK. Such improvement could improve and adoption of space cloud as most enterprise k8s clusters (and OpenShift by default) have such security requirements in place.

LECZYF:
fixes #1198

### DIFF
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -18,10 +18,18 @@ RUN echo $SC_VERSION && wget https://storage.googleapis.com/space-cloud/mission-
 FROM alpine:3.10
 ARG SC_VERSION=0.18.4
 
-RUN apk --no-cache add ca-certificates
+RUN apk --no-cache add ca-certificates && \
+    addgroup -S spacecloud && adduser -S spacecloud -G spacecloud && \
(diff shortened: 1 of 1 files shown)


YourTechBud:
Thanks for your contribution @LECZYF. Can you do the same for `runner` and `metric-proxy`?

LECZYF:
Sure, can do.

YourTechBud:
@LECZYF any progress on this?

izzatzr:
@YourTechBud Hi , i created PR #1564 for non root runner based on @LECZYF

PR #1205, opened 2020-08-06
Made changes according to the discussion

sharadregoti:


YourTechBud:
@sharadregoti close this if it isn't required

PR #1218, opened 2020-08-13
Tested team management console

sharadregoti:


gateway/modules/filestore/local/local.go
```diff
@@ -13,7 +14,10 @@ type Local struct {
 
 // Init initialises the local filestore driver
 func Init(path string) (*Local, error) {
-	return &Local{path}, nil
+	if !strings.HasPrefix(path, "/") {
```
YourTechBud, line 17: I think this is wrong. The rooth path can be relative too right?

sharadregoti: ok, then i'll remove this line

PR #1242, opened 2020-08-24
Added disable-ui flag in gateway

### LINKED ISSUE #1196: Create a flag to disable Mission Control
Currently gateway downloads and hosts mission control whether you want it or not. In many cases, a UI isn't required. Also, during local development, we host mission control locally.
### How can we solve it?
Create a flag `disable-ui` in gateway which stop SC from downloading or hosting mission control
>👍

KavishShah09:
Fixes #1196


YourTechBud (changes requested):
Dong add the handler for mission-control in routes.go

YourTechBud (approved):

PR #1285, opened 2020-09-04
Added gcs environment variable in kubernetes yaml file

### LINKED ISSUE #1250: support gcs setup on kubernetes with space cloud secrets
### The problem

Currently, space-cli sets the `GOOGLE_APPLICATION_CREDENTIALS=/root/.gcp/credentials.json` environment variable on the gateway container during setup so that we can use space cloud secrets to configure gcs file store. 

However, this won't work in the case of a Kubernetes setup, as space-cli isn't used there.

### How can we solve it?

Two possible ways:
- The gateway container itself sets the environment variable on start
- We specify the environment variable in the Kubernetes setup YAML file.

>👍

sharadregoti:
Fixes #1250

### DIFF
--- a/install-manifests/kubernetes/local/space-cloud.yaml
+++ b/install-manifests/kubernetes/local/space-cloud.yaml
@@ -388,6 +388,8 @@ spec:
                   fieldPath: metadata.name
             - name: "RUNNER_ADDR"
               value: "runner.space-cloud.svc.cluster.local:4050"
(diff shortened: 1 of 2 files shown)


YourTechBud:
@sharadregoti did you test this?

sharadregoti:
Yes

YourTechBud (approved):

PR #1286, opened 2020-09-05
Fields field in security rule takes values from args object

### LINKED ISSUE #1270: The fields section in remove, encrypt, decrypt and hash can be a variable
## The problem faced currently?
Currently the `fields` field in the security rule object can only be an array of string. There is no mechanism to load this from the `args` object.
### How can we solve it?
We should make the `fields` an `interface{}` instead of `[]string`.
>👍

sharadregoti:
Fixes #1270

### DIFF
--- a/gateway/modules/auth/match.go
+++ b/gateway/modules/auth/match.go
@@ -63,7 +63,7 @@ func (m *Module) matchRule(ctx context.Context, project string, rule *config.Rul
 		return m.matchDecrypt(ctx, rule, args)
 
 	case "hash":
-		return matchHash(ctx, rule, args)
+		return m.matchHash(ctx, rule, args)
 
 	default:
 		return nil, formatError(ctx, rule, fmt.Errorf("invalid rule type (%s) provided", rule.Rule))
@@ -189,7 +189,15 @@ func (m *Module) matchRemove(ctx context.Context, projectID string, rule *config
 		}
 	}
 	actions := &model.PostProcess{}
-	for _, field := range rule.Fields {
+	fields, err := m.getFields(ctx, rule.Fields, args)
+	if err != nil {
+		return nil, helpers.Logger.LogError(helpers.GetRequestID(ctx), "Invalid value provided for field (Fields) in security rules where rule is (Remove)", err, nil)
+	}
+	for _, value := range fields {
+		field, ok := value.(string)
+		if !ok {
+			return nil, helpers.Logger.LogError(helpers.GetRequestID(ctx), "Invalid value provided for field (Fields) in security rules where rule is (Remove) array contains a value which is not string", err, map[string]interface{}{})
+		}
 		// "res" - add field to structure for post processing || "args" - delete field from args
 		if strings.HasPrefix(field, "res") {
 			addToStruct := model.PostProcessAction{Action: "remove", Field: field, Value: nil}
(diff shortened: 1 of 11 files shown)


gateway/modules/auth/operations.go
```diff
@@ -12,3 +17,22 @@ func (m *Module) Encrypt(value string) (string, error) {
 
 	return utils.Encrypt(m.aesKey, value)
 }
+
+func (m *Module) getFields(ctx context.Context, fields interface{}, args map[string]interface{}) ([]interface{}, error) {
```
YourTechBud, line 21: Can we return array of string here?

YourTechBud: I suggest you move this function to `helpers.go`

YourTechBud (changes requested):
Resolve the conflicts as well

gateway/modules/auth/operations.go
sharadregoti: [] string? we can but an unnecessary for loop

sharadregoti: There is no helpers.go in auth package?

YourTechBud: So lets return [] interface only. Also, make a helpers.go file

gateway/modules/auth/match.go
```diff
@@ -201,7 +201,15 @@ func (m *Module) matchRemove(ctx context.Context, projectID string, rule *config
 		}
 	}
 	actions := &model.PostProcess{}
-	for _, field := range rule.Fields {
+	fields, err := m.getFields(ctx, rule.Fields, args)
+	if err != nil {
+		return nil, helpers.Logger.LogError(helpers.GetRequestID(ctx), "Invalid value provided for field (Fields) in security rules where rule is (Remove)", err, nil)
```
YourTechBud, line 206:
```suggestion
		return nil, err
```

gateway/modules/auth/match.go
```diff
@@ -228,7 +236,15 @@ func (m *Module) matchEncrypt(ctx context.Context, projectID string, rule *confi
 		}
 	}
 
-	for _, field := range rule.Fields {
+	fields, err := m.getFields(ctx, rule.Fields, args)
+	if err != nil {
+		return nil, helpers.Logger.LogError(helpers.GetRequestID(ctx), "Invalid value provided for field (Fields) in security rules where rule is (Encrypt)", err, nil)
```
YourTechBud, line 241:
```suggestion
		return nil, err
```

gateway/modules/auth/match.go
```diff
@@ -267,7 +282,15 @@ func (m *Module) matchDecrypt(ctx context.Context, projectID string, rule *confi
 		}
 	}
 
-	for _, field := range rule.Fields {
+	fields, err := m.getFields(ctx, rule.Fields, args)
+	if err != nil {
+		return nil, helpers.Logger.LogError(helpers.GetRequestID(ctx), "Invalid value provided for field (Fields) in security rules where rule is (Decrypt)", err, nil)
```
YourTechBud, line 287:
```suggestion
		return nil, err
```

gateway/modules/auth/match.go
```diff
@@ -321,7 +343,15 @@ func (m *Module) matchHash(ctx context.Context, projectID string, rule *config.R
 		}
 	}
 
-	for _, field := range rule.Fields {
+	fields, err := m.getFields(ctx, rule.Fields, args)
+	if err != nil {
+		return nil, helpers.Logger.LogError(helpers.GetRequestID(ctx), "Invalid value provided for field (Fields) in security rules where rule is (Hash)", err, nil)
```
YourTechBud, line 348:
```suggestion
		return nil, err
```

YourTechBud (changes requested):

YourTechBud (approved):

PR #1288, opened 2020-09-05
Store web hook & query result in args object

sharadregoti:


gateway/modules/auth/match.go
```diff
@@ -80,7 +81,18 @@ func (m *Module) matchFunc(ctx context.Context, rule *config.Rule, MakeHTTPReque
 	}
 
 	var result interface{}
-	return formatError(ctx, rule, MakeHTTPRequest(ctx, "POST", rule.URL, token, scToken, obj, &result))
+	if err := formatError(ctx, rule, MakeHTTPRequest(ctx, http.MethodPost, rule.URL, token, scToken, obj, &result)); err != nil {
```
YourTechBud, line 84: Call `formatError` only if `MakeHTTPRequest` returns an error

gateway/modules/auth/match.go
```diff
@@ -80,7 +81,18 @@ func (m *Module) matchFunc(ctx context.Context, rule *config.Rule, MakeHTTPReque
...
+		return err
+	}
+
+	if rule.Store != "" {
```
YourTechBud, line 88: If store value is not provided, store the result in `args.result` by default.

gateway/modules/auth/match.go
```diff
@@ -96,9 +108,14 @@ func (m *Module) matchQuery(ctx context.Context, project string, rule *config.Ru
...
-		return nil, formatError(ctx, rule, err)
+	if rule.Store != "" {
+		if err := utils.StoreValue(ctx, rule.Store, data, args); err != nil {
+			return nil, formatError(ctx, rule, err)
+		}
+	} else {
+		if err := utils.StoreValue(ctx, "args.result", data, args); err != nil {
+			return nil, formatError(ctx, rule, err)
+		}
 	}
```
YourTechBud, line 119: You can improve on the logic. 
Set the value of  `rule.Store` to `args.result` when empty.

YourTechBud (changes requested):
Let me know if you've tested this

sharadregoti:
> Let me know if you've tested this

Yes i did

YourTechBud (approved):

PR #1291, opened 2020-09-06
Added templating support in web hook security rules

sharadregoti:


gateway/modules/auth/match.go
```diff
@@ -72,8 +74,48 @@ func (m *Module) matchRule(ctx context.Context, project string, rule *config.Rul
...
+			t := template.New(rule.Name)
+			t = t.Funcs(tmpl2.CreateGoFuncMaps(m))
+			t, err = t.Parse(rule.ReqTmpl)
+			if err != nil {
+				return formatError(ctx, rule, helpers.Logger.LogError(helpers.GetRequestID(ctx), "Unable to parse provided template in security rule (Webhook)", err, nil))
+			}
+			if rule.OpFormat == "" {
+				rule.OpFormat = "json"
+			}
+			obj, err = tmpl2.GoTemplate(ctx, t, rule.OpFormat, token, claims, args["args"])
```
YourTechBud, line 108: put the original token and claims here

gateway/utils/tmpl/go.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 // GoTemplate executes a go template
-func GoTemplate(module, segment string, tmpl *template.Template, format, token string, claims, params interface{}) (interface{}, error) {
+func GoTemplate(ctx context.Context, tmpl *template.Template, format, token string, claims, params interface{}) (interface{}, error) {
```
YourTechBud, line 19: You took a context but are not using it. I suggest you use it instead of `context.Todo`

YourTechBud (changes requested):

gateway/modules/auth/match.go
sharadregoti: If I pass original token here, then what is the use of creating new token & claims if we are not using it?

YourTechBud: The point of claims is to generate a custom token to override the original token in the Authorisation header. This is useful when the webhook is on a SC endpoint and we want to make sure the request was made internally.

YourTechBud (approved):

PR #1303, opened 2020-09-14
Helm chart

kaivalya97:
Adding helm chart to deploy space cloud components

Usage: helm install space-cloud ./helm --atomic

Issue #1299

### DIFF
--- a/install-manifests/helm/Chart.yaml
+++ b/install-manifests/helm/Chart.yaml
@@ -1,0 +1,7 @@
+apiVersion: v2
+name: space-cloud
+description: Helm Chart to install Space Cloud
+type: application
+
+version: 0.1.0 # Chart version
+appVersion: 0.19.3 # Space Cloud version
--- a/install-manifests/helm/templates/space-cloud.yaml
+++ b/install-manifests/helm/templates/space-cloud.yaml
@@ -1,0 +1,428 @@
+
+#############################################################################################
+################################ Set some global parameters  ################################
+#############################################################################################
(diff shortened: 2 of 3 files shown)


helm/values.yaml
```diff
@@ -0,0 +1,24 @@
+# Default values for space-cloud.
+
+namespace: "space-cloud" # K8s namespace to deploy space cloud components
```
YourTechBud, line 3: We shouldn't keep this configurable. SC has a hard dependency on the namespace right now

helm/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+dev_mode: "true" # Turn this to false for production mode
+
+gateway:
+  image_name: "spaceuptech/gateway"
+  image_version: 0.19.2
+  replicas: 3
+  admin_user: "admin"
+  admin_password: "1234"
+  admin_secret: "some-secret" # Also used in runner as JWT_SECRET
```
YourTechBud, line 12: put these inside `admin` instead of `gateway`

helm/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+  cluster_id: "first-cluster" # Also used in runner
```
YourTechBud, line 13: This should be outside `gateway`

YourTechBud (changes requested):
Some changes
- Move `helm` dir to `install-manifests/helm`
- I've made some changes to the `space-cloud.yaml` file. Refer: https://github.com/spaceuptech/space-cloud/pull/1304

Issue #1071, opened 2020-06-11
Space cli should ask whether to put the binary in path


sharadregoti:
### The problem faced currently?
Most users access space-CLI binary by going to the downloaded path & then run commands. It would be nice if space-cli gives a promt of adding it to the path.
### How can we solve it?

>If you want this feature to be implemented, give it a thumbs up reaction, so that we can determine which features are important to you.
>👍

sharadregoti:
@YourTechBud what should we do about this issue ?

prakashmishra1598:
Rather than prompting the user, I think it would be nice to have a tiny installer script that does it. Something like a one-liner: `curl <url>/get_spacecloud.sh | sh` that could 
download and move the binary to /use/local/bin when executed. 

Also, the script could prompt the user if they want to add it to the path or just leave it in the current folder rather than implementing this logic in the binary. This approach just seems a bit cleaner to me. 

cc: @sharadregoti @YourTechBud

YourTechBud:
I agree.

sharadregoti:
yes, we should write a script for this. The script should also enable auto completion for cli

PR #1306, opened 2020-09-17
added space cli download script

sharadregoti:
I have to write another script for windows PowerShell


.gitignore
```diff
@@ -16,6 +16,7 @@ test
 
 # sh file for bash completion
 *.sh
+!install-manifests/cli/install.sh
```
YourTechBud, line 19: What is this for?

YourTechBud (changes requested):

.gitignore
sharadregoti: *.sh at line 18, means to exclude all the .sh files
As we now have an sh file, the (!) operator is used for adding some exceptions

YourTechBud (approved):

YourTechBud:
Fixes #1071

PR #1326, opened 2020-10-01
Create a new token if claims found in Endpoint Config

KhushalPShah:
- Added a Claims field in the Endpoint Config.
- If the Claims are set, then create a new token.


gateway/model/types.go
```diff
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	"github.com/spaceuptech/space-cloud/gateway/config"
+	"github.com/spaceuptech/space-cloud/gateway/model"
```
YourTechBud, line 8: No need to import this. You are already in the model package.

gateway/config/config.go
```diff
@@ -192,18 +192,19 @@ type Service struct {
...
-	Headers   Headers          `json:"headers,omitempty" yaml:"headers,omitempty"`
-	Timeout   int              `json:"timeout,omitempty" yaml:"timeout,omitempty"` // Timeout is in seconds
+	Kind      EndpointKind     		 `json:"kind" yaml:"kind"`
+	Tmpl      TemplatingEngine 		 `json:"template,omitempty" yaml:"template,omitempty"`
+	ReqTmpl   string           		 `json:"requestTemplate" yaml:"requestTemplate"`
+	GraphTmpl string           		 `json:"graphTemplate" yaml:"graphTemplate"`
+	ResTmpl   string           		 `json:"responseTemplate" yaml:"responseTemplate"`
+	OpFormat  string           		 `json:"outputFormat,omitempty" yaml:"outputFormat,omitempty"`
+	Token     string           		 `json:"token,omitempty" yaml:"token,omitempty"`
+	Claims    map[string]interface{} `json:"claims,omitempty" yaml:"claims,omitempty"`
```
YourTechBud, line 202: This file doesn't seem to be formatted right

gateway/model/types.go
```diff
@@ -65,6 +66,7 @@ type AuthFilestoreInterface interface {
 type AuthFunctionInterface interface {
 	GetSCAccessToken(ctx context.Context) (string, error)
 	Encrypt(value string) (string, error)
+	CreateToken(ctx context.Context, tokenClaims model.TokenClaims) (string, error)
```
YourTechBud, line 69:
```suggestion
	CreateToken(ctx context.Context, tokenClaims TokenClaims) (string, error)
```

YourTechBud (changes requested):

YourTechBud (approved):

PR #1362, opened 2020-10-13
Auth and DB resources deleters

### LINKED ISSUE #1155: Add a space-cli delete <resource> command
### How can we solve it?
>👍

KavishShah09:
fixes #1155 for auth and db module


space-cli/cmd/modules/auth/delete.go
```diff
@@ -0,0 +1,74 @@
...
+				return err
+			}
+		}
+	}
+
+	// Delete the provider from the server
+	url = fmt.Sprintf("/v1/config/projects/%s/user-management/provider/%s", project, prefix)
+
+	payload = new(model.Response)
+	if err := transport.Client.Get(http.MethodDelete, url, map[string]string{"id": prefix}, payload); err != nil {
```
sharadregoti, line 65: Rename the get function to makeHttpRequest

space-cli/cmd/modules/auth/delete.go
```diff
@@ -0,0 +1,74 @@
...
+		return err
+	}
+
+	if payload.Error != "" {
+		return utils.LogError(payload.Error, nil)
+	}
```
sharadregoti, line 71: Is this required ?, I think the Get function logs the error whenever the status code is 500

space-cli/cmd/modules/auth/delete.go
```diff
@@ -0,0 +1,74 @@
...
+	if err := transport.Client.Get(http.MethodGet, url, map[string]string{"id": "*"}, payload); err != nil {
+		return err
+	}
+
+	doesProviderExist := false
+	providers := []string{}
+	for _, item := range payload.Result {
+		spec := item.(map[string]interface{})
+		providers = append(providers, spec["id"].(string))
+	}
```
sharadregoti, line 30: Try to use the existing functions, GetAuthProviders which outputs spec object

sharadregoti (changes requested):

space-cli/cmd/modules/auth/delete.go
KavishShah09: But that is excess code of creating spec objects when only the id is required. Should I still use that function?

gateway/server/handlers/config_crud.go
```diff
@@ -504,6 +504,43 @@ func HandleGetTableRules(adminMan *admin.Manager, syncMan *syncman.Manager) http
...
+		vars := mux.Vars(r)
+		dbAlias := vars["dbAlias"]
+		projectID := vars["project"]
+		col := vars["col"]
+
+		v := config.TableRule{}
+		_ = json.NewDecoder(r.Body).Decode(&v)
+		defer utils.CloseTheCloser(r.Body)
+
+		ctx, cancel := context.WithTimeout(r.Context(), 60*time.Second)
```
sharadregoti, line 523: Use default context time out constant

space-cli/cmd/modules/delete.go
```diff
@@ -0,0 +1,18 @@
+package modules
+
+import (
+	"github.com/spaceuptech/space-cloud/space-cli/cmd/modules/auth"
+	"github.com/spf13/cobra"
+)
+
+// FetchDeleteSubCommands fetches all the delete subcommands from different modules
+func FetchDeleteSubCommands() *cobra.Command {
+	var generateCmd = &cobra.Command{
```
sharadregoti, line 10:
```suggestion
	var deleteCmd = &cobra.Command{
```

space-cli/cmd/modules/delete.go
```diff
@@ -0,0 +1,18 @@
...
+		Use:           "delete",
+		Short:         "",
+		SilenceErrors: true,
+	}
+	generateCmd.AddCommand(auth.DeleteSubCommands()...)
+
```
sharadregoti, line 16: You haven't added the delete DB-rule sub command here

space-cli/cmd/modules/auth/commands.go
```diff
@@ -81,3 +81,46 @@ func actionGenerateUserManagement(cmd *cobra.Command, args []string) error {
...
+			objs, err := GetAuthProviders(project, "auth-providers", map[string]string{})
+			if err != nil {
+				return nil, cobra.ShellCompDirectiveDefault
+			}
+			var ids []string
+			for _, v := range objs {
+				ids = append(ids, v.Meta["id"])
+			}
+			return ids, cobra.ShellCompDirectiveDefault
+		},
```
sharadregoti, line 107: I think this piece of code is also written in get auth-provider command.
I suggest you make a file called helpers.go & define this function in that file & use this function in getter & delete subcommand.
Do the same for other commands also, DB-rules, DB-config etc...
Keep the code DRY

space-cli/cmd/modules/auth/delete.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// Delete the provider from the server
+	url := fmt.Sprintf("/v1/config/projects/%s/user-management/provider/%s", project, prefix)
+
+	payload := new(model.Response)
+	if err := transport.Client.MakeHTTPRequest(http.MethodDelete, url, map[string]string{"id": prefix}, payload); err != nil {
+		return err
+	}
+
+	return nil
```
sharadregoti, line 65: This piece of code is also getting repeated a lot, can you try making a generic function which does this. And call that function everywhere

sharadregoti (changes requested):

space-cli/cmd/modules/database/helpers.go
```diff
@@ -0,0 +1,132 @@
+package database
+
+import (
+	"github.com/spaceuptech/space-cloud/space-cli/cmd/utils"
+	"github.com/spf13/cobra"
+)
+
+func validDBRulesArgsFunc(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
sharadregoti, line 8:
```suggestion
func dbRulesAutoCompleteFun(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
do the same with other functions in this file

sharadregoti (changes requested):

space-cli/cmd/modules/database/commands.go
```diff
@@ -348,3 +230,127 @@ func actionGenerateDBPreparedQuery(cmd *cobra.Command, args []string) error {
...
+		RunE:              actionDeleteDbPreparedQuery,
+		ValidArgsFunction: dbPreparedQueriesAutoCompleteFun,
+	}
+
+	var deleteSchemas = &cobra.Command{
+		Use:               "db-prepared-queries",
+		Aliases:           []string{"db-prepared-query"},
+		RunE:              actionDeleteDbSchemas,
+		ValidArgsFunction: dbSchemasAutoCompleteFun,
+	}
```
sharadregoti, line 263: Specify the usage of these commands

sharadregoti (changes requested):

sharadregoti (approved):

PR #1378, opened 2020-10-23
Filestore space cli deleters

### LINKED ISSUE #1155: Add a space-cli delete <resource> command

KavishShah09:
fixes #1155 for filestore module


gateway/managers/syncman/syncman_file.go
```diff
@@ -36,6 +36,31 @@ func (s *Manager) SetFileStore(ctx context.Context, project string, value *confi
...
+		return http.StatusInternalServerError, helpers.Logger.LogError(helpers.GetRequestID(ctx), "error setting file store config", err, nil)
+	}
+
+	resourceID := config.GenerateResourceID(s.clusterID, project, config.ResourceFileStoreConfig, "filestore")
+	if err := s.store.DeleteResource(ctx, resourceID); err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	return http.StatusOK, nil
+}
```
sharadregoti, line 62: I don't think so this is required, you just call the set file store config endpoint with all the required values in the payload as empty

space-cli/cmd/modules/filestore/commands.go
```diff
@@ -60,6 +45,27 @@ func GetSubCommands() []*cobra.Command {
...
 
+// DeleteSubCommands is the list of commands the filestore module exposes
+func DeleteSubCommands() []*cobra.Command {
+
+	var deleteFileStoreRules = &cobra.Command{
+		Use:               "filestore-rules",
+		Aliases:           []string{"filestore-rule"},
+		RunE:              actionDeleteFileStoreRule,
+		ValidArgsFunction: fileStoreRulesAutoCompleteFun,
+		Example:           "space-cli delete filestore-rules ruleID --project myproject --log-level info",
```
sharadregoti, line 56: Remove --log level info from usage.
Do this everywhere

space-cli/cmd/utils/filter/filter.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 // DeleteOptions filters the resource id based on prefix to delete
-func DeleteOptions(project, prefix string, resources []string, doesResourceExist bool) (string, error) {
+func DeleteOptions(prefix string, resources []string, doesResourceExist bool) (string, error) {
```
sharadregoti, line 12: chnage line number 15 to
		if prefix != "" && strings.HasPrefix(strings.ToLower(resource), strings.ToLower(prefix)) && prefix != resource{

And remove the if at line number 22, always show the filtered resources

sharadregoti (changes requested):

KavishShah09:
@sharadregoti please add `hacktoberfest-accepted` label to this

sharadregoti (approved):

PR #1379, opened 2020-10-23
Ingress space-cli deleters

### LINKED ISSUE #1155: Add a space-cli delete <resource> command

KavishShah09:
fixes #1155 for ingress module


gateway/managers/syncman/syncman_routing.go
```diff
@@ -174,6 +174,33 @@ func (s *Manager) SetGlobalRouteConfig(ctx context.Context, project string, glob
...
+	s.modules.Routing().SetGlobalConfig(projectConfig.IngressGlobal)
+
+	// Finally lets store the config
+	resourceID := config.GenerateResourceID(s.clusterID, project, config.ResourceIngressGlobal, "global")
+	if err := s.store.DeleteResource(ctx, resourceID); err != nil {
+		return http.StatusInternalServerError, err
+	}
+
+	return http.StatusOK, nil
+}
```
sharadregoti, line 202: Not required, use setter endpoints to set empty values

sharadregoti (changes requested):

space-cli/cmd/modules/ingress/commands.go
```diff
@@ -126,3 +111,48 @@ func actionGetIngressGlobal(cmd *cobra.Command, args []string) error {
...
+
+// DeleteSubCommands is the list of commands the ingress module exposes
+func DeleteSubCommands() []*cobra.Command {
+
+	var deleteRoutes = &cobra.Command{
+		Use:               "ingress-routes",
+		Aliases:           []string{"ingress-route"},
+		RunE:              actionDeleteIngressRoutes,
+		ValidArgsFunction: ingressRoutesAutoCompleteFun,
+		Example:           "space-cli delete ingress-routes routeID --project myproject --log-level info",
```
sharadregoti, line 123: remove --log-level info, from all commands

space-cli/cmd/modules/ingress/delete.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func deleteIngressRoute(project, prefix string) error {
+
+	objs, err := GetIngressRoutes(project, "ingress-route", map[string]string{}, []string{})
+	if err != nil {
+		return err
+	}
+
+	doesRouteExist := false
```
sharadregoti, line 32: What is the use of this boolean?

space-cli/cmd/modules/ingress/delete.go
```diff
@@ -0,0 +1,52 @@
...
+	routeIDs := []string{}
+	for _, spec := range objs {
+		routeIDs = append(routeIDs, spec.Meta["id"])
+	}
+
+	prefix, err = filter.DeleteOptions(prefix, routeIDs, doesRouteExist)
```
sharadregoti, line 38: prefix sounds misleading, call it resourceID
Do this every where

space-cli/cmd/modules/ingress/delete.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/spaceuptech/space-cloud/space-cli/cmd/utils/transport"
+)
+
+func deleteIngressGlobalConfig(project string) error {
+
+	// Delete the ingress global config from the server
+	url := fmt.Sprintf("/v1/config/projects/%s/routing/ingress/global", project)
+
+	payload := new(model.Response)
+	if err := transport.Client.MakeHTTPRequest(http.MethodPost, url, map[string]string{}, payload); err != nil {
```
sharadregoti, line 18:
```suggestion
	if err := transport.Client.MakeHTTPRequest(http.MethodPost, url, map[string]string{}, new(model.Response)); err != nil {
```
Do this everywhere, where you are not using the payload

sharadregoti (changes requested):

sharadregoti:
Test this once, you don't need Kubernetes for this

space-cli/cmd/modules/ingress/delete.go
KavishShah09: Used in differentiating when to survey only filtered resource IDs and when to show all

sharadregoti: You don't need to pass this as a parameter to the function. Just create the boolean locally inside the Deleteoptions function.

The boolean that you are passing as a parameter is always false

sharadregoti (approved):

Issue #476, opened 2019-11-02
[RFC] Schema SDL for primary key


HeyItsJs:
## The `@primary` directive

The `@primary` directive is used currently to make a particular field as primary key. However it has the following limitations:
- No support for composite primary key
- Only a field of type `ID` can be made primary.

I suggest the following SDL for the `@primary` directive:
- Should be applicable for following field types:
  - ID
  - Integer
- Arguments:
  - strategy: ENUM (optional):
    - AUTO (default): Will generate random unique string values for type `ID` and auto-increment/serial values for type `Integer`. If the user provides the value for this field in mutation, then the value from user will be used.
    - NONE: Will not generate any value. User will have to provide value during mutation.
- If the `@primary` directive is used on multiple fields in a type, then those fields should form a composite primary key

_Examples_ 

Composite primary key:
```graphql
type transactions {
  transaction_id: ID! @primary
  batch_id: ID! @primary
  date: DateTime!
  payload: String
}
```

In the above example, `transaction_id` and `batch_id` will together form a composite primary key.

Integer primary key:
```graphql
type orders {
  order_no: Integer! @primary
  amount: Float!
  date: DateTime!
}
```

Disabling auto-generation of primary key:
```graphql
type orders {
  order_id: ID! @primary(strategy: "NONE")
  amount: Float!
  date: DateTime!
}
```

JoeyJanson:
@sharadregoti @YourTechBud Will this be assigned on a new milestone?

HeyItsJs:
Hey @JoeyJanson, which part of this are you interested in? The _composite_ key part or supporting other types for primary key such as auto-increment integer?

JoeyJanson:
> Hey @JoeyJanson, which part of this are you interested in? The _composite_ key part or supporting other types for primary key such as auto-increment integer?

Other types for primary key. It was the first issue I ran in to after connecting with an existing MySQL database.

HeyItsJs:
It can be done for sure and definitely on our roadmap. Not sure whether it can be done in this release itself. @YourTechBud and @sharadregoti  would be able to tell that better.

YourTechBud:
@JoeyJanson which datatype are you looking for? 
@HeyItsJs it will be very difficult to add support for auto incrementing keys. Our `linked` inserts feature will break in that case.

JoeyJanson:
[...]

PR #1393, opened 2020-11-08
Add missing database types

sharadregoti:
1) Type UUID support for Postgres
2) Type Time & Date Support for all the SQL databases
3) Fixed default value problem
4) Now primary keys can be created with type Integer
5) Auto Increment of Primary keys

NOTE: 
1) Inspection for auto-increment for SQL server does not work, I was not able to find the correct DB query for that.
2) Date & Time types have not been tested for db read operation


YourTechBud:
Fixes #476

gateway/modules/crud/sql/describe.go
```diff
@@ -50,6 +54,10 @@ CASE
     WHEN t.constraint_type = 'UNIQUE' THEN 'UNI'
     ELSE ''
 END AS "Key",
+CASE 
+	WHEN isc.data_type = 'bigint' AND t.constraint_type = 'PRIMARY KEY' THEN 'true'
```
YourTechBud, line 58: What if someone had used serial?

gateway/modules/schema/inspection.go
```diff
@@ -78,15 +76,6 @@ func generateInspection(dbType, col string, fields []model.InspectorFieldType, f
...
 
-			if model.DBType(dbType) == model.Postgres {
-				// split "'default-value'::text" to "default-value"
-				s := strings.Split(field.FieldDefault, "::")
-				field.FieldDefault = s[0]
-				if fieldDetails.Kind == model.TypeString || fieldDetails.Kind == model.TypeDateTime || fieldDetails.Kind == model.TypeID {
-					field.FieldDefault = strings.Split(field.FieldDefault, "'")[1]
-				}
-			}
-
```
YourTechBud, line 89: Are you sure we can remove this?

gateway/modules/schema/inspection.go
```diff
@@ -64,11 +64,9 @@ func generateInspection(dbType, col string, fields []model.InspectorFieldType, f
 		}
 
 		// default key
-		if field.FieldDefault != "" {
+		if field.FieldDefault != "" && field.AutoIncrement == "false" {
 			fieldDetails.IsDefault = true
 			if model.DBType(dbType) == model.SQLServer {
-				// replace (( or )) with nothing e.g -> ((9.8)) -> 9.8
-				field.FieldDefault = strings.Replace(strings.Replace(field.FieldDefault, "(", "", -1), ")", "", -1)
```
YourTechBud, line 71: Are you sure about this?

YourTechBud (changes requested):

gateway/modules/crud/sql/describe.go
sharadregoti: Altaf has sent me a new query for Postgres & SQL server. I have to test that out

gateway/modules/schema/inspection.go
sharadregoti: Yes, To confirm you can check the Postgres query in describe table, The filtering logic is implemented in the query itself

gateway/modules/schema/inspection.go
sharadregoti: Yes, To confirm you can check the SQL Server query in describe table, The filtering logic is implemented in the query itself

gateway/modules/crud/sql/describe.go
YourTechBud: Okay. So I won't merge this PR till then

gateway/modules/schema/inspection.go
```diff
@@ -99,6 +88,18 @@ func generateInspection(dbType, col string, fields []model.InspectorFieldType, f
 			fieldDetails.IsPrimary = true
 		}
 
+		// Set auto increment
+		fieldDetails.AutoIncrementInfo = new(model.AutoIncrementInfo)
+		if field.AutoIncrement == "true" {
+			fieldDetails.AutoIncrementInfo = &model.AutoIncrementInfo{IsEnabled: true}
```
YourTechBud, line 94: What about start from and increment by?

YourTechBud (changes requested):

gateway/modules/schema/inspection.go
YourTechBud: I would suggest you remove support for increment by and start from since its available in sqlserver only. In that case this piece of code is right. You'll have to change the part where you parse the schema

YourTechBud (approved):

PR #1396, opened 2020-11-10
WIP: Services Deleters

### LINKED ISSUE #1155: Add a space-cli delete <resource> command

KavishShah09:
fixes #1155 for services module


KavishShah09:
Won't be able to work on this till after December due to some reasons

PR #1443, opened 2020-12-15
Service request retries & timeouts

### LINKED ISSUE #1173: Let user configure request retries and timeouts for requests going to services deployed by SC
We internally configure Istio to set automatic retries and timeouts for requests going to each service. The user might want to override those defaults
### How can we solve it?
Allow users to override these settings. We can have a section in the `deployments` config to accept this defaults. We will have to take care to use defaults for backward compatibilty.

>👍

### LINKED ISSUE #1437: support op type one in graphql db reads
GraphQL request to query a table always returns an array. There is no option to make it return an object
### How can we solve it?
Add an op clause to the request to return a single object when needed. By default the op will be always to maintain backward compatibility.
```graphql
{
 table(op: one) @db
}
```

**We need to make sure native joins can work when doing op: one as well.**
>👍

sharadregoti:
1) Graphql now supports op argument to return an only single result, Fixes #1437 
2) Added the ability to specify request retries & timeout for services deployed through space cloud, Fixes #1173

### DIFF
--- a/gateway/modules/crud/dataloader.go
+++ b/gateway/modules/crud/dataloader.go
@@ -13,8 +13,13 @@ import (
 
 type resultsHolder struct {
 	sync.Mutex
-	results      []*dataloader.Result
-	whereClauses []interface{}
+	results []*dataloader.Result
+	metas   []meta
+}
+
+type meta struct {
+	whereClause map[string]interface{}
+	op          string
 }
 
 func (holder *resultsHolder) getResults() []*dataloader.Result {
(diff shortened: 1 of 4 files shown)


gateway/modules/crud/dataloader.go
```diff
@@ -67,20 +76,33 @@ func (holder *resultsHolder) fillResults(res []interface{}) {
...
 			}
+			if isOperationTypeOne {
+				break
+			}
 		}
 
 		// Increment the where clause index
 		index++
 
+		if isOperationTypeOne {
```
YourTechBud, line 95: Make the result variable here

gateway/modules/crud/dataloader.go
```diff
@@ -67,20 +76,33 @@ func (holder *resultsHolder) fillResults(res []interface{}) {
...
+			if len(docs) >= 1 {
```
YourTechBud, line 96:
```suggestion
			if len(docs) > 0 {
```

gateway/modules/crud/dataloader.go
```diff
@@ -67,20 +76,33 @@ func (holder *resultsHolder) fillResults(res []interface{}) {
...
+				result = docs[0]
+			} else {
+				result = map[string]interface{}{}
```
YourTechBud, line 99: Maybe you should throw an error here or make it nil

YourTechBud (changes requested):
Are you sure Space cloud has support for tci services? Did you try it?

YourTechBud (approved):

PR #1465, opened 2020-12-24
Data type fixes

sharadregoti:
1) Now you can specify tablename.columnname syntax in mongodb where clause. space cloud will remove the tablename prefix depending upon the collection being queried.
2) In & not in operator now works with boolean data type
3) Now we can use logical operators on datetime data type in mongodb
4) Mysql will now use decimal data type for float you can specify the precision & scale in the schema
5) Fixed DateTime data type in mysql not storing milli/nano seconds.
6) Fixed Time data type not storing mill/nano seconds


gateway/model/schema_type.go
```diff
@@ -15,6 +15,12 @@ type (
 		IsFieldTypeRequired bool   `json:"isFieldTypeRequired"`
 		IsList              bool   `json:"isList"`
 		Kind                string `json:"kind"`
+		// Precision is used to hold precision information for data types Float
+		// It represent number the digits to be stored
+		Precision int `json:"precision"`
```
YourTechBud, line 20: Store such things inside an args struct

gateway/modules/crud/mgo/helpers_test.go
```diff
@@ -0,0 +1,85 @@
...
+		args args
+		want map[string]interface{}
+	}{
+		{
+			name: "Where clause with collection name separated by dot notation",
+			args: args{
+				ctx: context.Background(),
+				col: "users",
+				find: map[string]interface{}{
+					"users.id":         1,
```
YourTechBud, line 26: Add one more field, users.address.street. the result should be address.street

gateway/modules/crud/mgo/read.go
```diff
@@ -24,6 +24,8 @@ func (m *Mongo) Read(ctx context.Context, col string, req *model.ReadRequest) (i
 	}
 	collection := m.getClient().Database(m.dbName).Collection(col)
 
+	req.Find = sanitizeWhereClause(ctx, col, req.Find)
```
YourTechBud, line 27: Are you doing this for update and delete as well?

YourTechBud (changes requested):

gateway/modules/crud/mgo/read.go
sharadregoti: Nope, I'll add it

YourTechBud (approved):

PR #1468, opened 2020-12-26
Fixed SQL inspection queries

sharadregoti:
1) Fixed indexes getting detected with specific prefix(index__), now space cloud can detect index starting from any name. Support has been added only for (Mysql,Postgres)
2) Added inspection support for @args directive, following are fields are detected (precision & scale)


sharadregoti:
Following tasks are remaining

1. Add schema inspection for composite primary keys
2. Add schema inspection for multiple unique/index on a single column
3. Update describe table SQL queries for SQL Server Database

sharadregoti:
Following tasks are remaining

- Update inspection SQL server query to change data type to json from nvarchar if check_json constraint is fount on column
- Update describe queries for postgres & SQL server to incorporate database other than the default

sharadregoti:
All the tasks have been completed

YourTechBud (approved):

PR #1487, opened 2021-01-02
Made dataloader case insensitive

sharadregoti:
1) Made data loader case insensitive for mysql & sql server

### DIFF
--- a/gateway/modules/crud/operations.go
+++ b/gateway/modules/crud/operations.go
@@ -65,7 +65,11 @@ func (m *Module) Read(ctx context.Context, dbAlias, col string, req *model.ReadR
 	}
 
 	if req.IsBatch {
-		key := model.ReadRequestKey{DBType: dbAlias, Col: col, HasOptions: req.Options.HasOptions, Req: *req, ReqParams: params}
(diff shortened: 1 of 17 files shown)


gateway/modules/crud/operations.go
```diff
@@ -65,7 +65,11 @@ func (m *Module) Read(ctx context.Context, dbAlias, col string, req *model.ReadR
 	}
 
 	if req.IsBatch {
-		key := model.ReadRequestKey{DBType: dbAlias, Col: col, HasOptions: req.Options.HasOptions, Req: *req, ReqParams: params}
+		dbType, err := m.GetDBType(dbAlias)
```
YourTechBud, line 68: Does this acquire a read lock?

YourTechBud (changes requested):

YourTechBud (approved):

gateway/modules/crud/operations.go
YourTechBud: Dont export it if it doesn't acquire a lock

YourTechBud (approved):

PR #1495, opened 2021-01-09
Modified helper package to include send empty raw error

sharadregoti:


sharadregoti:
Following changes were done:

- Modified helper package to include send empty raw error
- Fixed, unable to track  Postgres table with a column having data type serial with no primary key on it
- Sqlserver & Postgres now use decimal data type for Float, now you can set the precision and scale using @args directive
- Removed constraint of having primary on the table.
- Fixed, gateway crashing when a prepared query is queried

YourTechBud (approved):

PR #1497, opened 2021-01-12
Removed auto generate ID functionality if the field is nullable.

### LINKED ISSUE #1496: Support for Varchar field

The type `ID` in schema gets auto-generated even if the field is nullable. This is not desired for optional fields. Rather we can have a type `Varchar` with `@size` directive that never auto-generates values

### How can we solve it?

>👍

sharadregoti:
1) From now onwards, If a field in a schema has a type of ID which is not null. Then space cloud will not auto-generate ID for that field. For auto-generate to work, you have make the field required by adding (!) in the schema.

Fixes #1496

### DIFF
--- a/gateway/utils/graphql/create.go
+++ b/gateway/utils/graphql/create.go
@@ -44,7 +44,7 @@ func (graph *Module) prepareDocs(doc map[string]interface{}, schemaFields model.
 	fieldDefaults := make(map[string]interface{})
 
 	for fieldName, fieldSchema := range schemaFields {
-		// ignore nullable id fields
+		// Only process ID fields which are required
 		if fieldSchema.Kind == model.TypeID && fieldSchema.IsFieldTypeRequired {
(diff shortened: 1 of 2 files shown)


gateway/utils/graphql/create.go
```diff
@@ -44,7 +44,8 @@ func (graph *Module) prepareDocs(doc map[string]interface{}, schemaFields model.
 	fieldDefaults := make(map[string]interface{})
 
 	for fieldName, fieldSchema := range schemaFields {
-		if fieldSchema.Kind == model.TypeID {
+		// ignore nullable id fields
```
YourTechBud, line 47:
```suggestion
		// Only process ID fields which are required
```

YourTechBud (changes requested):
Is this tested with GraphQL and HTTP api?

YourTechBud (approved):

PR #1512, opened 2021-01-30
Added graphql & transform security rule

### LINKED ISSUE #1478: Add ability to disable eventing on specific databases and tables.
Debezium gets overwhelmed when we enable eventing on a project with multiple databases.
### How can we solve it?
Add ability to optionally disable eventing altogether on selected database and tables. By default eventing will be on for everything.
>👍

sharadregoti:
This fixes #1478 #1482 #1483 
Following changes were done …
1) Added support to enable eventing on specific tables & databases.
2) Added graphQL security rule.
3) Added transform security rule.

TODO
- [ ] Update the database filtering logic & database connectors in the scala codebase to support enabling eventing on specific tables and databases

### DIFF
--- a/gateway/modules/auth/match.go
+++ b/gateway/modules/auth/match.go
@@ -74,6 +74,12 @@ func (m *Module) matchRule(ctx context.Context, project string, rule *config.Rul
 	case "hash":
 		return m.matchHash(ctx, project, rule, args, auth)
 
+	case "graphql":
+		return m.matchGraphQL(ctx, project, rule, args, auth, returnWhere)
+
+	case "transform":
+		return m.matchTransform(ctx, rule, args)
+
 	default:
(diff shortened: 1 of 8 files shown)


gateway/modules/auth/match.go
```diff
@@ -437,3 +443,79 @@ func (m *Module) matchHash(ctx context.Context, projectID string, rule *config.R
...
+	}
+
+	type result struct {
+		op  interface{}
+		err error
+	}
+	ch := make(chan result, 1)
+	defer close(ch)
+
+	m.graphql.ExecGraphQLQuery(ctx, &model.GraphQLRequest{Query: rule.GraphQLQuery, Variables: rule.GraphQLVariables}, token, func(op interface{}, err error) {
```
YourTechBud, line 471: The variables need to be adjusted. Check the query `where` clause for example.

YourTechBud (changes requested):

PR #1579, opened 2021-08-28
v0.21.5

sharadregoti:


install-manifests/docker/mongo/docker-compose.yaml
```diff
@@ -1,7 +1,7 @@
 version: "3.9"
 services:
   gateway:
-    image: "spaceuptech/gateway:0.21.4"
+    image: "spaceuptech/gateway:0.21.5"
```
YourTechBud, line 4: The image name will change to `spacecloudio/gateway`. Do this in kubernetes files and helm chart too

YourTechBud (changes requested):

YourTechBud (approved):

PR #1594, opened 2022-01-28
fix schema inspect

junfengye:
remove break inside for loop


YourTechBud (dismissed):
Ooo. Not sure how we missed that one.

YourTechBud (approved):

PR #1595, opened 2022-01-29
fix enpoint pointer issue

junfengye:
directly modify endpoint.Headers will cause header data duplication next time we update the service config


YourTechBud (approved):
Totally missed that once. Thanks for the contribution

YourTechBud (changes requested):
Simple change to please the linter

PR #1597, opened 2022-02-06
support db and password config for redis connection

junfengye:
parse REDIS_CONN with redis.ParseURL in order to support db and password config for redis connection


YourTechBud (approved):
Wow. You have really gone ahead deep into the code for this one. Thanks for your contribution.

PR #1599, opened 2022-02-19
Add env variables

khannaabhi:


sc/cmd/space-cli/commands/commands.go
```diff
@@ -1,26 +1,24 @@
...
 )
 
-func GetRootCommand() *cobra.Command {
-	rootCmd := &cobra.Command{
+// NewRootCommand returns space-cli command
+func NewRootCommand() *cobra.Command {
+	cmd := &cobra.Command{
 		Use:          "space-cli",
 		Version:      "v0.22.0",
 		SilenceUsage: true,
```
YourTechBud, line 14: Do we have to do this

.github/workflows/go.yml
```diff
@@ -32,6 +32,37 @@ jobs:
           sudo cp ./bin/golangci-lint $GOPATH/bin/
           cd gateway && golangci-lint run -E golint --exclude-use-default=false --timeout 600s ./...
 
+  caddy-space-cli:
+    name: Build and test caddy-space-cli
```
YourTechBud, line 36: Lets add a paths field to prevent all jobs from running

sc/utils/config.go
```diff
@@ -2,39 +2,54 @@ package utils
...
 )
 
 // LoadAdminConfig creates a caddy config from the viper config provided. This only contains the admin
 // and logging portion of the configuration. The config loaders will be responsible to load the
 // configuration of the applications.
-func LoadAdminConfig(isInitialLoad bool) *caddy.Config {
-	// TODO: Read base config from viper over here.
-	loadingInterval := caddy.Duration(time.Second) * 5
+func LoadAdminConfig(isInitialLoad bool) (*caddy.Config, error) {
+	interval := viper.GetInt64("loading-interval")
```
YourTechBud, line 16: I suggest you accept a string here which is in the duration format

sc/utils/config.go
```diff
@@ -2,39 +2,54 @@ package utils
...
-				LoadRaw: prepareFileLoaderConfig(),
+				LoadRaw:      loader,
 			},
 		},
-		// TODO: Configure logging as well
-	}
+		Logging: &caddy.Logging{
+			Logs: map[string]*caddy.CustomLog{
+				"default": {
+					Level: viper.GetString("log-level"),
```
YourTechBud, line 42: I suggest we create variables for each one of the input flags at the beginning of the function so we know what all parameters are being used.

YourTechBud (changes requested):

PR #1647, opened 2023-02-08
Remove baseUrl from flag and instead provide a new file to configure baseUrl

### LINKED ISSUE #1646: [Feature] Take fetchBaseQuery config from a file instead of cli

Developers need control over baseUrl in client gen. So the config can be taken from a file which is generated only if it doesn't exists.

### How can we solve it?

Remove the baseUrl flag and create a new file for the config.

>👍

shubham4443:
Signed-off-by: Shubham Nazare <shubham4443@gmail.com>

Fixes #1646

### DIFF
--- a/sc-poc/cmd/spacectl/commands/client/generate/config.go
+++ b/sc-poc/cmd/spacectl/commands/client/generate/config.go
@@ -2,6 +2,6 @@ package generate
 
 var configTS string = `
 export default {
-	"baseUrl": "/",
+  "baseUrl": "/",
 }
 `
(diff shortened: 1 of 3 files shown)


sc-poc/cmd/spacectl/commands/client/generate/config.go
```diff
@@ -0,0 +1,7 @@
+package generate
+
+var configTS string = `
+export default {
+	"baseUrl": "/",
```
YourTechBud, line 5:
```suggestion
  "baseUrl": "/",
```

YourTechBud (approved):

PR #1670, opened 2023-04-16
feat!: add websocket interface for pubsub

### LINKED ISSUE #1668: [Feature] Create websocket manager
- Create websocket manager for our pubsub
- It acts as an interface between clients and the pubsub system.
- The protocol makes it possible for messages to be published and subscribed.

shubham4443:
Fixes #1668
The PR adds the ability to perform publish/subscribe operations with the websocket protocol.

### DIFF
--- a/sc-poc/modules/pubsub/types.go
+++ b/sc-poc/modules/pubsub/types.go
@@ -1,5 +1,20 @@
 package pubsub
 
+type (
+	eventType string
+)
+
+const (
+	subscribeEvent eventType = "subscribe"
+	messageEvent   eventType = "message"
+)
+
+// Message defines the type of event and the associated data
+type Message struct {
+	Event eventType              `json:"event"`
+	Data  map[string]interface{} `json:"data"`
+}
+
 // PublishMessage defines the type for publishing a message
 type PublishMessage struct {
 	ID       string            `json:"id"`
@@ -14,6 +29,10 @@ type PublishOptions struct {
 
 // SubscribeOptions defines the options for subscribing a message
 type SubscribeOptions struct {
+	Mode     string `json:"mode"`
+	Capacity int    `json:"capacity"`
+	Autoack  bool   `json:"autoack"`
+	Format   string `json:"format"`
 }
 
 // ChannelsWithSchema define the channels schema and component
--- a/sc-poc/modules/pubsub/types.go
+++ b/sc-poc/modules/pubsub/types.go
@@ -5,8 +5,9 @@ type (
 )
 
 const (
-	subscribeEvent eventType = "subscribe"
-	messageEvent   eventType = "message"
+	subscribeEvent   eventType = "subscribe"
+	unsubscribeEvent eventType = "unsubscribe"
+	messageEvent     eventType = "message"
 )
 
 // Message defines the type of event and the associated data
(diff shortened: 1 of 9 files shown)


sc-poc/modules/pubsub/websocket.go
```diff
@@ -0,0 +1,165 @@
...
+
+	"github.com/spacecloud-io/space-cloud/managers/apis"
+)
+
+// TODO: channels may or may not have prefix slash
+
+// GetRoutes returns all the apis that are exposed by this app
+func (a *App) GetAPIRoutes() apis.APIs {
+	channels := a.Channels()
+	apis := []*apis.API{}
```
YourTechBud, line 23: I suggest we make this api in `Provision` and simply return it here

sc-poc/modules/pubsub/websocket.go
```diff
@@ -0,0 +1,165 @@
...
+			// Create the websocket connection
+			conn, _, _, err := ws.UpgradeHTTP(r, w)
+			if err != nil {
+				a.logger.Error("could not establish websocket connection", zap.String("channel", channelName), zap.Error(err))
+				return
+			}
+
+			// Go routine to receive messages from the websocket connection and publish
+			// them into the watermill channel
+			go func() {
```
YourTechBud, line 51: I don't think we need to run this as a seperate goroutine.

You can confirm this by starting a subscriber then trying to reload the config. If the config gets reloaded without affecting the socket we are good.

sc-poc/modules/pubsub/websocket.go
```diff
@@ -0,0 +1,165 @@
...
+					if message.Event == messageEvent {
+						var pubMsg PublishMessage
+						err = mapstructure.Decode(message.Data, &pubMsg)
+						if err != nil {
+							a.logger.Error("could not decode data", zap.String("channel", channelName), zap.Error(err))
+							continue
+						}
+
+						if pubMsg.ID == "" {
+							pubMsg.ID = uuid.NewString()
```
YourTechBud, line 77: Use ksuid

sc-poc/modules/pubsub/websocket.go
```diff
@@ -0,0 +1,165 @@
...
+
+			// Create the websocket connection
+			conn, _, _, err := ws.UpgradeHTTP(r, w)
+			if err != nil {
+				a.logger.Error("could not establish websocket connection", zap.String("channel", channelName), zap.Error(err))
+				return
+			}
+
+			// Wait for subscribe event
+			for {
```
YourTechBud, line 108: This implementation won't work cause it doesn't leave room for you to check for the unsubscribe event that could come in next

YourTechBud (changes requested):

sc-poc/modules/pubsub/websocket.go
shubham4443: Should we use a variable which stores the state(subscribed/not subscribed) in which the client is in right now?

sc-poc/modules/pubsub/types.go
```diff
@@ -1,5 +1,21 @@
 package pubsub
```
YourTechBud, line 2: Export `eventType` and the various constants

YourTechBud (changes requested):

PR #1698, opened 2023-06-29
Create endpoint to return available plugins and create 2 additional plugins

### LINKED ISSUE #1689: [Feature] Add endpoint to return a list of available plugins.

There is no way to figure out how many plugins are available/ registered in SC.

### How can we solve it?

- Each `Source` which feeds into a plugin (eg. `OPAPolicySource`) should implement a `source.Plugin` interface.
- The interface will be defined by the `source` manager.
- The `Plugin` interface will have just one method, `GetPluginDetails()` which returns a struct containing `name`, `driver` and `schema`.
    - Name will be the name of the source
    - Driver will be the handler which needs to be initialised for that plugin (for `OPAPolicySource`, the driver will be `opa`).
    - Params will always be `nil` for now.
- The `Source` manager should expose an api `/sc/v1/plugins` which returns array of plugins registered.

>👍

### LINKED ISSUE #1692: [Feature] Create handlers for checking if session is valid
Need to make two handlers in the `auth` provider and `admin` manager respectively
- `deny-user`
    - This plugin immediately sends a `403`.
    - Implement this in the `auth` provider.
    - We need to make sure that one entry for this plugin (name will be empty string) is added in the plugin list that we return in #1689.
- `authenticate-user`
    - This will merely check if the user session is valid by checking the `types.AuthResult` object present in the context
        - Return `401` if user is not authenticated
        - For authResult reference use: https://github.com/spacecloud-io/space-cloud/blob/sc-poc/sc-poc/modules/auth/handler_auth_opa.go#L46
    - We need to make sure that one entry for this plugin (name will be empty string) is added in the plugin list that we return in #1689.
- `authenticate-sc-user`
    - This checks if user is authenticated for by using the admin secret. Send `403` if token is not present or if token is invalid.
    - Consider making this handler in the admin module
    - This handler will have to perform a jwt validation using the admin secret provided in #1686 
    - This plugin will not be returned in the plugin details list.

Also, make sure the following sc apis use the `authenticate-sc-user` plugin:
- All CRUD APIs
- the get crds api
- The pubsub consumer channels for asyncapi and openapi. The producer channels will have the `deny-user` plugin
- The get plugin details api

>👍

shubham4443:
Fixes #1689 
Partially fixes #1692 

Requires https://github.com/spacecloud-io/space-cloud/pull/1697 to be merged first.


sc-poc/managers/source/app.go
```diff
@@ -49,6 +51,22 @@ func (a *App) Provision(ctx caddy.Context) error {
 				continue
 			}
 
+			a.plugins = make([]v1alpha1.HTTPPlugin, 0)
```
YourTechBud, line 54: Shouldn't you do this outside all the for loops

YourTechBud (changes requested):

YourTechBud (approved):

